PR #3, opened 2017-06-14
Remove superfluous private clause

esparta:
This last `private` is not needed since we it was added before.

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -2321,8 +2321,6 @@ class CSV
     chunks.map { |chunk| chunk.encode(@encoding.name) }.join('')
   end
 
-  private
-
   #
   # Returns the encoding of the internal IO object or the +default+ if the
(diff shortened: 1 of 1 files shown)


amatsuda:
Thanks!

esparta:
No, thank you!
I found this trying to split this atom. Is there any technical reason to have all the lib on a single file instead of multiple ones?

amatsuda:
Well, just because that's the way how this program was originally written: https://github.com/JEG2/faster_csv/blob/master/lib/faster_csv.rb

esparta:
Ok. Is good to know.
Later on I'll propose to split on at least 3 files (core, CSV::Table and CSV::Row). The lib is now almost 2.5K lines of code, and IMO can be more manageable (for humans) on three 900+ lines per file.

Thank you again.

PR #4, opened 2017-06-30
Add the ability to delete multiple columns

wowinter13:
When I was working with some statistics using Ruby/csv, I discovered that the mechanism for deleting columns is very peculiar in Ruby and we can't easily just delete more than 1 column.
We can do it like this (DRY):
```ruby
csv_table.delete("PassengerId")
csv_table.delete("Survived")
csv_table.delete("Name")
csv_table.to_csv
```

Or like this (will destroy csv structure -> Array of Arrays):
```ruby
datus = CSV.readlines('titanic.csv', headers:true).map{|row| row.values_at('PassengerId','SibSp') }
```

And finally the best working way for me:
```ruby
to_del = ["Name", "Ticket"]
csv_table = datus.by_col!.delete_if { |name, values| !to_del.include? name}
```
But I think it should work easily like this:
```ruby
to_del = ["Name", "Ticket"]
data.delete(to_del)
```
So I rewrote method a little and added this functionality. If you have any ideas: create another ```delete_columns``` method or maybe better ways to fast delete columns, I will be glad to hear it:)

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -807,20 +807,18 @@ class CSV
     end
 
     #
-    # Removes and returns the indicated columns or row.  In the default mixed
+    # Removes and returns the indicated columns or rows.  In the default mixed
     # mode indices refer to rows and everything else is assumed to be a column
-    # header.  Use by_col!() or by_row!() to force the lookup.
+    # headers.  Use by_col!() or by_row!() to force the lookup.
     #
-    def delete(index_or_header)
-      if index_or_header.is_a? Array # by array of headers
-        index_or_header.each do |element|
-          @table.map { |row| row.delete(element) }
+    def delete(*indexes_or_headers)
+      indexes_or_headers.map do |index_or_header|
+        if @mode == :row or  # by index
+          (@mode == :col_or_row and index_or_header.is_a? Integer)
+          @table.delete_at(index_or_header)
+        else                 # by header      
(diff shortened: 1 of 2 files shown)


lib/csv.rb
```diff
@@ -807,15 +807,19 @@ def push(*rows)
     end
 
     #
-    # Removes and returns the indicated column or row.  In the default mixed
+    # Removes and returns the indicated columns or row.  In the default mixed
     # mode indices refer to rows and everything else is assumed to be a column
     # header.  Use by_col!() or by_row!() to force the lookup.
     #
     def delete(index_or_header)
```
kou: `*index_or_header` is better than accepting an `Array` argument.

lib/csv.rb
```diff
@@ -807,15 +807,19 @@ def push(*rows)
...
-      if @mode == :row or  # by index
+      if index_or_header.is_a? Array # by array of headers
+        index_or_header.each do |element|
```
kou: This method should return the deleted objects.

```ruby
indexes_or_headers.map do |index_or_header|
  delete(index_or_header)
end
```

wowinter13:
@kou Thanks for review and I changed checked methods. Can you view current version please?

kou:
Thanks.
I've merged and reverted zero argument (raising `ArgumentError`) and one argument (returning without wrapping by an `Array`) case behavior for backward compatibility.

PR #11, opened 2017-12-14
Add proper description to gemspec

stomar:
This adds a proper description, taken from the README, for displaying on rubygems.org (the historic origins of the library are not too helpful there).

_Besides, I would also suggest to change the repo description from ""the CSV library began its life as FasterCSV" to "CSV Reading and Writing" (the summary in the gemspec)._


hsbt:
Thanks! I updated repository description too.

PR #13, opened 2017-12-28
fix generate with a string

ahorek:
ruby 2.5 regression

https://github.com/ruby/ruby/commit/a983fc502288776cde7b97519c7ddab7aa2ca6b9

expected
```ruby
CSV.generate("test") { |csv| csv << ["row"] }
=> testrow\n
```

actual
```ruby
CSV.generate("test") { |csv| csv << ["row"] }
=> row\n
```

http://ruby-doc.org/stdlib-2.5.0/libdoc/csv/rdoc/CSV.html#method-c-generate

> This method wraps a String you provide, or an empty default String, in a CSV object which is passed to the provided block. You can use the block to append CSV rows to the String and when the block exits, the final String will be returned.

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1164,8 +1164,8 @@ class CSV
   def self.generate(str=nil, **options)
     # add a default empty String, if none was given
     if str
-      io = StringIO.new(str)
-      io.seek(0, IO::SEEK_END)
+      str = StringIO.new(str)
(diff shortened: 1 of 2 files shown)


hsbt:
At least, Please show your expected and actual behavior.

kou:
Good catch!

Issue #5, opened 2017-07-12
Proposal: Split csv.rb into a more manageable structure


esparta:
Currently `CSV` is a single file library, being a 2300+ LOC file is kind of difficult to manage by itself. We probably need to follow the [Rubygem's proposal][rubygem] on how to structure the gem.
Pros:
- Explore, check, review and maintain 5 to 6 files of 500+ LOC is (based on my own experience) more easy to manage
- Next commits can and should affect relevant class, making the git log more user friendly.
- Reduce and distribute the churn across history of the gem

Cons:
- At the moment none, at least no technical

Plan:
- Extract internal classes:
  - [x] CSV::Table -> /lib/csv/table.rb
  - [ ] CSV::Row -> /lib/csv/row.rb
- Move extensions to their own namspace/files
  - [X] Array -> /lib/core_ext/array.rb
  - [X] String -> /lib/core_ext/string.rb

With this we will have 5 files, and still work to do via future refactors (a 1700+ LOC file still looks like a red flag) and unit test.

[rubygem]: http://guides.rubygems.org/make-your-own-gem/#requiring-more-files

stevendaniels:
While I agree that `csv` could benefit from structure changes, we should be careful not to make too many changes too quickly.

It's also worth noting that the `csv.rb` **is only**  ~1000 LOC without comments.

stevendaniels:
@esparta - This has been completed and this issue can be closed.

esparta:
Thanks @stevendaniels, @hsbt & @kou to make this happen!

PR #14, opened 2017-12-28
Move extensions to their own namespaced files

stevendaniels:
Implements part of https://github.com/ruby/csv/issues/5

This does not change any behavior of the csv gem.

@hsbt - is there a preference to using require/require_relative here?


stevendaniels:
@hsbt  - I renamed `ext` to `core_ext`.

kou:
Thanks!

PR #15, opened 2018-01-12
[Feature] Add dig method to CSV::Table and CSV::Row

tomog105:
## Abstract
- This PR implements `dig` method to CSV::Table and CSV::Row

## Background
- Already, Array, Hash and Struct class has `dig` method from Ruby 2.3, but CSV::Table and CSV::Row not has this method.
- I would like to use `#dig` for CSV::Table or nested arrays (or hashes) with CSV::Row without distinction.

## Usage

```rb
arr_of_arrs = CSV.parse("first,second,third\n1,2,3\n", headers: true)

# as single value
arr_or_arrs.dig(0) # row mode
=> #<CSV::Row "first":"1" "second":"2" "third":"3">
arr_of_arrs.dig("first") # col mode
=> ["1"]

# as nested values
arr_of_arrs.dig(0, "first")
=> "1"
arr_of_arrs.dig("first", 0)
=> "1"
```

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -548,7 +548,7 @@ class CSV
     alias_method :to_s, :to_csv
 
     #
-    # Extracts the nested value specified by the sequence of name objects by calling dig at each step,
+    # Extracts the nested value specified by the sequence of +index+ or +header+ objects by calling dig at each step,
     # returning nil if any intermediate step is nil.
     #
     def dig(index_or_header, *indexes)
@@ -923,7 +923,7 @@ class CSV
     alias_method :to_s, :to_csv
 
     #
-    # Extracts the nested value specified by the sequence of name objects by calling dig at each step,
+    # Extracts the nested value specified by the sequence of +index+ or +header+ objects by calling dig at each step,
     # returning nil if any intermediate step is nil.
     #
     def dig(index_or_header, *index_or_headers)
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -547,6 +547,21 @@ class CSV
     end
     alias_method :to_s, :to_csv
 
+    #
+    # Extracts the nested value specified by the sequence of name objects by calling dig at each step,
+    # returning nil if any intermediate step is nil.
+    #
+    def dig(index_or_header, *index_or_headers)
+      value = field(index_or_header)
+      if value.nil?
+        nil
+      elsif index_or_headers.empty?
+        value
+      else
+        value.dig(*index_or_headers)
+      end
+    end
+
     # A summary of fields, by header, in an ASCII compatible String.
     def inspect
       str = ["#<", self.class.to_s]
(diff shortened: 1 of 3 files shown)


lib/csv.rb
```diff
@@ -903,6 +918,21 @@ def to_csv(write_headers: true, **options)
     end
     alias_method :to_s, :to_csv
 
+    #
+    # Extracts the nested value specified by the sequence of name objects by calling dig at each step,
+    # returning nil if any intermediate step is nil.
+    #
+    def dig(index_or_header, *index_or_headers)
+      value = send(:[], index_or_header)
```
kou: `self[index_or_header` is better than `send(:[], index_or_header)`.

tomog105: Thanks, I fixed this part.

test/csv/test_row.rb
```diff
@@ -377,4 +377,16 @@ def test_can_be_compared_when_not_a_row
...
+  def test_dig
+    # by index
+    assert_equal(2, @row.dig(1))
+
+    # by header
+    assert_equal(2, @row.dig("B"))
+
+    # if missing
+    assert_nil(@row.dig("Missing"))
+    assert_nil(@row.dig(100))
```
kou: Can you add a test for multiple arguments case?

lib/csv.rb
```diff
@@ -547,6 +547,21 @@ def to_csv(**options)
     end
     alias_method :to_s, :to_csv
 
+    #
+    # Extracts the nested value specified by the sequence of name objects by calling dig at each step,
+    # returning nil if any intermediate step is nil.
+    #
+    def dig(index_or_header, *index_or_headers)
```
kou: `*index_or_headers` isn't a good name because they are arguments for `Array#dig`. `Array` doesn't have header.
`*indexes` is better.

lib/csv.rb
```diff
@@ -547,6 +547,21 @@ def to_csv(**options)
     end
     alias_method :to_s, :to_csv
 
+    #
+    # Extracts the nested value specified by the sequence of name objects by calling dig at each step,
```
kou: I don't think that `the sequence of name objects` is good description.
`the +index+ or +header+ object and the followed sequence of +index+ objects` (this English may be broken) or something will be better.

lib/csv.rb
```diff
@@ -903,6 +918,21 @@ def to_csv(write_headers: true, **options)
...
```
kou: Same here.

test/csv/test_table.rb
```diff
@@ -494,4 +494,49 @@ def test_inspect_encoding_is_ascii_compatible
...
+    ### Row Mode ###
+    ################
+    @table.by_row!
+
+    assert_equal(@rows[1], @table.dig(1))
+    assert_raise(TypeError) { @table.dig("B") }
+
+    # by cell, row then col
+    assert_equal(2, @table.dig(0, 1))
+    assert_equal(6, @table.dig(1, "C"))
```
kou: Can you add tests for three or more arguments case?

lib/csv.rb
tomog105: Thanks, I corrected arguments name.

test/csv/test_row.rb
tomog105: I added multiple arguments case.

test/csv/test_table.rb
tomog105: I added three or more arguments case.

tomog105:
I fixed about the part reviewed.
and I changed behavior when following value does not have `#dig` method.
(would raise same error as Array and Hash `#dig`.)

kou:
Thanks.
I've merged with some modifications.

PR #18, opened 2018-01-30
improve by Ruby 2.5 Lasy Proc

takkanm:
Ruby2.5 has the advantage to use `block_given?`.

https://bugs.ruby-lang.org/issues/14045

```
$ ruby -v
ruby 2.5.0p0 (2017-12-25 revision 61468) [x86_64-darwin17]

$ cat block_given.rb
def use_if_block(&block)
  return nil if block
end

def use_block_given(&block)
  return nil if block_given?
end

require 'benchmark'
Benchmark.bm(10){|x|
  x.report("use_if_block"){
    100000.times{
      use_if_block { 1 }
    }
  }
  x.report("use_block_given") {
    100000.times{
      use_block_given { 1 }
    }
  }
}

$ ruby block_given.rb
                 user     system      total        real
use_if_block  0.053502   0.001040   0.054542 (  0.054878)
use_block_given  0.010321   0.000023   0.010344 (  0.010386)
```

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1310,8 +1310,9 @@ class CSV
   def self.parse(*args, &block)
     csv = new(*args)
 
-    return csv.each(&block) if block_given? # slurp contents, if no block is given
+    return csv.each(&block) if block_given?
 
+    # slurp contents, if no block is given
     begin
       csv.read
     ensure
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -839,7 +839,7 @@ class CSV
       else                                      # by header
         deleted = []
         headers.each do |header|
-          deleted << delete(header) if block[[header, self[header]]]
+          deleted << delete(header) if yield([header, self[header]])
         end
       end
 
@@ -861,7 +861,7 @@ class CSV
       return enum_for(__method__) { @mode == :col ? headers.size : size } unless block_given?
 
       if @mode == :col
-        headers.each { |header| block[[header, self[header]]] }
+        headers.each { |header| yield([header, self[header]]) }
       else
         @table.each(&block)
       end
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -434,7 +434,7 @@ class CSV
     # If no block is given, an Enumerator is returned.
     #
     def delete_if(&block)
-      block or return enum_for(__method__) { size }
+      return enum_for(__method__) { size } unless block_given?
 
       @row.delete_if(&block)
 
@@ -513,7 +513,7 @@ class CSV
     # Support for Enumerable.
     #
     def each(&block)
-      block or return enum_for(__method__) { size }
+      return enum_for(__method__) { size } unless block_given?
 
       @row.each(&block)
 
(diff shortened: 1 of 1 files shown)


lib/csv.rb
```diff
@@ -1309,7 +1309,7 @@ def self.open(filename, mode="r", **options)
   #
   def self.parse(*args, &block)
     csv = new(*args)
-    if block.nil?  # slurp contents, if no block is given
+    unless block_given? # slurp contents, if no block is given
```
kou: `unless else` is confused.
How about the following?

```ruby
return csv.each(&block) if block_given?

begin
  csv.read
ensure
  csv.close
end
```

kou:
I'll merge this later.
BTW, do you have any benchmark with this change? I think that we need more changes to use `yield` instead of `block[...]` to get the Ruby 2.5 improvement.

lib/csv.rb
takkanm: I'll try it!

takkanm:
> BTW, do you have any benchmark with this change? 

No. I'm getting benchmark. But, anytime fast, anytime slow.

> I think that we need more changes to use yield instead of block[...] to get the Ruby 2.5 improvement.

I think too, and I already have a commit.  But, I am at a loss as to whether to mix this pull request.

lib/csv.rb
```diff
@@ -1309,14 +1309,13 @@ def self.open(filename, mode="r", **options)
...
-    if block.nil?  # slurp contents, if no block is given
-      begin
-        csv.read
-      ensure
-        csv.close
-      end
-    else           # or pass each row to a provided block
-      csv.each(&block)
+
+    return csv.each(&block) if block_given? # slurp contents, if no block is given
```
kou: The comment should be placed at the following code block not this sentence.

kou:
Can you show the benchmark scripts you're using?

> I am at a loss as to whether to mix this pull request.

They should be included in this pull request because the Ruby 2.5 improvement can't be used without the changes.

takkanm:
> Can you show the benchmark scripts you're using?

I use this code.

https://gist.github.com/takkanm/8a62cf54ceff71f71bfabddddf78039b

kou:
Thanks.
The benchmark is meaningless because it doesn't use block.

kou:
I confirmed that this change doesn't effect performance by the following script:

```ruby
require "benchmark"
require "csv"

def CSV.parse2(*args, &block)
  csv = new(*args)

  return csv.each(&block) if block_given?

  # slurp contents, if no block is given
  begin
    csv.read
  ensure
    csv.close
  end
end

csv_text = (0..10).map { (?a..?z).to_a.join(",") }.join("\n")

GC.disable
Benchmark.bmbm(10){|x|
  x.report("before"){
    100.times{
      CSV.parse(csv_text) do
      end
    }
  }
  x.report("after"){
    100.times{
      CSV.parse2(csv_text) do
      end
    }
  }
}
```

Result:

```text
Rehearsal ----------------------------------------------
before       0.044303   0.000000   0.044303 (  0.044303)
after        0.043426   0.000000   0.043426 (  0.043427)
------------------------------------- total: 0.087729sec

                 user     system      total        real
before       0.043628   0.000000   0.043628 (  0.043626)
after        0.044570   0.000000   0.044570 (  0.044568)
```

I'll merge this change just for code clean up not for performance improvement.

takkanm:
Thank you for your support.

PR #19, opened 2018-02-10
Moving CSV::Row to their own file

esparta:
Implements part of https://github.com/ruby/csv/issues/5

This does not change any behavior of the csv gem.


lib/csv.rb
```diff
@@ -589,6 +215,8 @@ def inspect
   # All tables returned by CSV will be constructed from this class, if header
   # row processing is activated.
   #
+
+  require_relative "csv/row"
```
kou: We should put `require` family at the top level not in `class`.

lib/csv/row.rb
```diff
@@ -0,0 +1,384 @@
+# frozen_string_literal: true
+# = csv.rb -- CSV Reading and Writing
```
kou: This file isn't `csv.rb`.
We can remove the header description by moving the description to README.

esparta:
Done with the changes requested. Thanks, @kou !

lib/csv.rb
```diff
@@ -589,6 +216,7 @@ def inspect
   # All tables returned by CSV will be constructed from this class, if header
   # row processing is activated.
   #
+
```
stevendaniels: Should we remove this extra line-break?

kou: @esparta Can you remove this change?

esparta: sure, no problem

kou:
Thanks!

PR #20, opened 2018-02-11
Move CSV::Table to a separate file

stevendaniels:
+ Moved the CSV::Table class to a separate file

This PR Implements part of #5.

It does not change any behavior of the csv gem.


kou:
Thanks!

PR #22, opened 2018-02-12
Update Test Features Herdoc

stevendaniels:
Although the current syntax works, it's not very "readable".

This does not change any functionality.


kou:
Thanks!!!

PR #24, opened 2018-02-16
Fix Unknown Encoding issue

### LINKED ISSUE #23: Regression since ruby 2.4 - `unknown encoding name - bom|utf-8`
Hi there I update my app to ruby-2.5 and have some problems with sending encoding `bom|utf-8` to csv parser.

I create a sample Dockerfiles for this:
With `ruby-2.4.3` everything working fine:
```
FROM ruby:2.4.3

RUN echo 'test' > test.csv
RUN echo "require 'csv'\ncsv = CSV.read('test.csv', encoding: 'bom|utf-8')\n p csv[0][0]" > script.rb
CMD ruby script.rb
```
output is `"test"`

But for ruby-2.5.0 encoding error is happend
```
FROM ruby:2.5.0

RUN echo 'test' > test.csv
RUN echo "require 'csv'\ncsv = CSV.read('test.csv', encoding: 'bom|utf-8')\n p csv[0][0]" > script.rb
CMD ruby script.rb
```

Error is:
```
/usr/local/lib/ruby/2.5.0/csv.rb:1532:in `find': unknown encoding name - bom|utf-8 (ArgumentError)
        from /usr/local/lib/ruby/2.5.0/csv.rb:1532:in `initialize'
        from /usr/local/lib/ruby/2.5.0/csv.rb:1280:in `new'
        from /usr/local/lib/ruby/2.5.0/csv.rb:1280:in `open'
        from /usr/local/lib/ruby/2.5.0/csv.rb:1346:in `read'
        from script.rb:2:in `<main>'
```

stevendaniels:
This fixes #23 and makes the following modification:

+ @encoding can be set to external_encoding. Previously, it ignored external_encoding.

To me, it looks like the encoding option is usually ignored. `CSV.read|CSV.open` will use the encoding options provided by a user if:

1. the user passes in an `IO`-like object where `respond_to?(:internal_encoding) == true && (io.internal_encoding || io.external_encoding).nil?`
2. the user passes in a `StringIO`-like object where the string has no encoding.
3. the user passes in an `IO`-like object where`respond_to?(:encoding) == true && io.encoding.nil?`

As far as I know, the above situations can only occur if the user is passing in a custom `IO` object
that doesn't have encoding information. This analysis is based off of the `raw_encoding` method

```ruby
def raw_encoding(default = Encoding::ASCII_8BIT)
  if @io.respond_to? :internal_encoding
    @io.internal_encoding || @io.external_encoding
  elsif @io.is_a? StringIO
    @io.string.encoding
  elsif @io.respond_to? :encoding
    @io.encoding
  else
    default
  end
end
```

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1569,15 +1569,10 @@ class CSV
 
     # create the IO object we will read from
     @io = data.is_a?(String) ? StringIO.new(data) : data
-    # honor the IO encoding if we can, otherwise default to ASCII-8BIT
-    internal_encoding = Encoding.find(internal_encoding) if internal_encoding
-    external_encoding = Encoding.find(external_encoding) if external_encoding
-    if encoding
-      encoding, = encoding.split(":", 2) if encoding.is_a?(String)
-      encoding = Encoding.find(encoding)
-    end
-    @encoding = raw_encoding(nil) || internal_encoding || encoding ||
-                Encoding.default_internal || Encoding.default_external
+
+    # gets the encoding we will use
+    @encoding = get_encoding(internal_encoding, external_encoding, encoding)
+
     #
     # prepare for building safe regular expressions in the target encoding,
     # if we can transcode the needed characters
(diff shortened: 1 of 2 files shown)


kou:
Thanks for your work but I've fixed it.
Because we need more new separated tests for this case instead of changing existing tests.

PR #28, opened 2018-03-06
Remove redundant `private` method

ShockwaveNN:
There is already `private` on https://github.com/ruby/csv/blob/e0a25dbb268ad78d6c0f9f995cb05a8ef00437fc/test/lib/test/unit.rb#L558


kou:
Good catch!

PR #32, opened 2018-03-25
Update CVS documentation

zverok:
Problems of the "intro" section of [current documentation](http://ruby-doc.org/stdlib-2.5.0/libdoc/csv/rdoc/CSV.html) I am trying to solve:

* It does not provide a generic overview of the class usage, diving straight into particular examples;
* The overall structure, with a lot of headers and subheaders, which separate 1-2 line sections, is hard to read
* The large part of functionality (data conversion and usage of headers to obtain `CSV::Table`/`CSV::Row`) is not mentioned in the intro at all. I constantly see students, juniors or even middle devs that use the library a lot yet still unaware of those parts of functionality.

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -101,68 +101,117 @@ require_relative "csv/row"
 # tools to enable you to read and write to and from Strings or IO objects, as
 # needed.
 #
-# == Reading
+# The most generic interface of a class is:
 #
-# === From a File
+#    csv = CSV.new(string_or_io, **options)
 #
-# ==== A Line at a Time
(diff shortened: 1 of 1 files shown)


kou:
Thanks!

lib/csv.rb
```diff
@@ -101,68 +101,117 @@
 # tools to enable you to read and write to and from Strings or IO objects, as
 # needed.
 #
-# == Reading
+# The most generic interface of a class is:
```
stomar: @zverok unclear to me: „of a class“ (?)

PR #33, opened 2018-03-25
Add Row#each_pair

zverok:
Reason: `Row` tries to quack more-or-less like `Hash`, and `Hash` has `#each_pair`. A practical example where it matters:

```ruby
csv = CSV.parse('Bob,100', headers: %i[name salary])
OpenStruct.new(csv.first)
# NoMethodError: undefined method `each_pair' for #<CSV::Row name:"Bob" salary:"100">

# After this patch:
OpenStruct.new(csv.first)
#=> #<OpenStruct name="Bob", salary="100">
```

### DIFF
--- a/lib/csv/row.rb
+++ b/lib/csv/row.rb
@@ -313,6 +313,8 @@ class CSV
       self  # for chaining
     end
 
+    alias each_pair each
+
     #
     # Returns +true+ if this row contains the same headers and fields in the
     # same order as +other+.


kou:
Do you want to try adding a test?

zverok:
Thought about it, but I am not sure what is an appropriate amount of testing. It is just an alias, so do you expect me to copy-paste of tests of `#each`, or just literally test "the alias, named this, exists"?

zverok:
> What is the expected behavior of the following case?

On my branch:

```ruby
csv.first.each_pair.to_a
# => [[:name, "Bob"], [:salary, "100"]]
```

Is there some problem?..

kou:
Ah. it's not related with this change.
It's my fault.

PR #40, opened 2018-07-03
Fix typo

stomar:
@kou Not sure what the customs are in this repo; if you prefer I could commit minor changes like this directly.


kou:
Thanks!
You can commit directly for trivial fixes like we did for the Ruby's Subversion repository.

PR #41, opened 2018-07-04
Add descriptions of new options. (nil_value and empty_value)

owlworks:
An description of the new options on Rdoc is blank, so I tried adding it.


kou:
Thanks!

Issue #44, opened 2018-09-08
Reading from a non-rewindable stream that supports pos can cause row_sep autodetection to silently corrupt data


vyznev:
As the title says, passing an IO object that supports `pos` but not `rewind` to `CSV.new` or `CSV.parse` can cause silent data corruption (specifically, loss of the first 1024 bytes of input) if `row_sep` is not explicitly specified.

I originally ran into this problem on JRuby, while trying to use a non-markable Java InputStream (specifically, a `java.util.zip.ZipInputStream`) wrapped with `.to_io` as the input to `CSV.parse`.  However, it's also possible to [demonstrate the problem](https://tio.run/##TVDLboMwELz7K7ZwAKTIaq6RKlVqVSmnHCL1VCly8Ia6BS@1TWkKfDtdSJRy8D5mdsdju@Z4HkeHX41xCEnuvxNx63xwxhaGFtCJXKucVscSEyHyUnkPz01Vnbc7AYA/Aa2Gl/8hBjWeDhpLLFQg52HzaGgFmwKD5@RQaU41eYAYLIFH/ARy4LA1Vt9d9sFYE4wqzS@mWgWVMQzAQvAA@9nkdicttjeSXYjpCH4Qz6RrKdf3maxUDZqgd9T2k8QFvxId9DmVPUTMxh2Hgdu44zBEMMhAByU/yNg0WkXZpL6E3ixj4mn/Otu4/shcs4Msk6jy9@XVdRM8cC2N9TXmYdIbxz8 "Ruby – Try It Online") with a simple wrapper around StringIO:

```
require 'csv'
require 'stringio'
require 'forwardable'

class DummyIO
  extend Forwardable
  def_delegators :@io, :gets, :read, :pos  # no seek or rewind!
  def initialize(data)
    @io = StringIO.new(data)
  end
end

csv = (1..10).map do |row|
   (1..100).map { |col| "row#{row}col#{col}" }.to_a.join(",")
end.to_a.join("\n")

CSV.new(DummyIO.new(csv)).each do |row|
  puts row.inspect
end
```

The output of this code, which *should* begin with `["row1col1", "row1col2", "row1col3"`, instead begins with `["ol4", "row2col5", "row2col6", "row2col7"`.

Ideally, the `row_sep` autodetection should be rewritten to save the data it reads, so that it can be later parsed without having to seek or rewind the input at all. This would allow any non-seekable input streams to be safely used without needing any special-case hacks.

vyznev:
[...]

PR #45, opened 2018-09-11
Cache input data read during init_separators to avoid seek/rewind.

vyznev:
Here's a basic patch that avoids calling seek or rewind after row_sep autodetection (https://github.com/ruby/csv/issues/44) by saving the sample data and wrapping it in a StringIO object for re-reading. It passes all existing tests.

It could also be a good idea to add some new tests, e.g. using a dummy IO wrapper like in my [online demonstration](https://tio.run/##TVDLboMwELz7K7ZwAKTIaq6RKlVqVSmnHCL1VCly8Ia6BS@1TWkKfDtdSJRy8D5mdsdju@Z4HkeHX41xCEnuvxNx63xwxhaGFtCJXKucVscSEyHyUnkPz01Vnbc7AYA/Aa2Gl/8hBjWeDhpLLFQg52HzaGgFmwKD5@RQaU41eYAYLIFH/ARy4LA1Vt9d9sFYE4wqzS@mWgWVMQzAQvAA@9nkdicttjeSXYjpCH4Qz6RrKdf3maxUDZqgd9T2k8QFvxId9DmVPUTMxh2Hgdu44zBEMMhAByU/yNg0WkXZpL6E3ixj4mn/Otu4/shcs4Msk6jy9@XVdRM8cC2N9TXmYdIbxz8), to verify that the new implementation works properly with non-seekable input streams, but I have not done that yet in this patch.

Note that this patch also removes several special cases from init_separators. In particular, row_sep autodetection should now work when reading from a pipe or from ARGF or STDIN. This should probably also have some tests added for it.

I have not run any benchmarks to compare this new implementation with the previous one, but I suspect that they should be pretty close. In some cases the new code might even be faster, e.g. if (re)reading the input data is really slow for some reason.

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -912,6 +912,7 @@ class CSV
 
     # create the IO object we will read from
     @io = data.is_a?(String) ? StringIO.new(data) : data
+    @prefix_io = nil  # cache for input data possibly read by init_separators
     @encoding = determine_encoding(encoding, internal_encoding)
     #
     # prepare for building safe regular expressions in the target encoding,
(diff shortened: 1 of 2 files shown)


kou:
I'll merge this after tests for this changes are added.
Can you work on adding tests?

kou:
Thanks.
I've merged this.

Test for pipe isn't needed because it's covered by the `DummyIO` case.

PR #47, opened 2018-10-25
Enable headers-only file to get headers using 'headers: true' option

shibasaka:
When trying to get headers from headers-only file using `headers: true` option, Table#headers method currently returns [].
So, I thought it's proper that Table#headers method return file's header in this case.

For example:
test.csv(headers-only file)
```
a,b,c
```

the previous behaviour
```
csv = CSV.read 'test.csv', {headers: true}
csv.headers
=> []
```

changed it as follows
```
csv = CSV.read 'test.csv', {headers: true}
csv.headers
=> ["a", "b", "c"]
```

I added an argument of headers to `Table.new` method because Table class should always has headers.

### DIFF
--- a/lib/csv/table.rb
+++ b/lib/csv/table.rb
@@ -23,7 +23,8 @@ class CSV
     # * length()
     # * size()
     #
-    def initialize(array_of_rows)
+    def initialize(array_of_rows, headers=nil)
+      @headers = headers || []
       @table = array_of_rows
       @mode  = :col_or_row
     end
@@ -122,11 +123,7 @@ class CSV
     # other rows).  An empty Array is returned for empty tables.
     #
     def headers
-      if @table.empty?
-        Array.new
-      else
-        @table.first.headers
-      end
+      @headers.dup
     end
 
     #
@@ -171,6 +168,10 @@ class CSV
           @table[index_or_header] = value
         end
       else                 # set column
+        unless index_or_header.is_a? Integer
+          index = @headers.index(index_or_header) || @headers.size
+          @headers[index] = index_or_header
+        end
         if value.is_a? Array  # multiple values
           @table.each_with_index do |row, i|
             if row.header_row?
@@ -258,6 +259,11 @@ class CSV
             (@mode == :col_or_row and index_or_header.is_a? Integer)
           @table.delete_at(index_or_header)
         else                 # by header
+          if index_or_header.is_a? Integer
+            @headers.delete_at(index_or_header)
+          else
+            @headers.delete(index_or_header)
+          end
           @table.map { |row| row.delete(index_or_header).last }
         end
       end
@@ -285,7 +291,8 @@ class CSV
         @table.delete_if(&block)
       else                                      # by header
         deleted = []
-        headers.each do |header|
+        initial_headers = headers.dup
+        initial_headers.each do |header|
           deleted << delete(header) if yield([header, self[header]])
         end
       end
--- a/lib/csv/table.rb
+++ b/lib/csv/table.rb
@@ -16,6 +16,11 @@ class CSV
     # Construct a new CSV::Table from +array_of_rows+, which are expected
     # to be CSV::Row objects.  All rows are assumed to have the same headers.
     #
+    # The optional +headers+ parameter can be set to Array of headers.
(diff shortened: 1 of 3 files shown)


lib/csv/table.rb
```diff
@@ -258,6 +259,11 @@ def delete(*indexes_or_headers)
             (@mode == :col_or_row and index_or_header.is_a? Integer)
           @table.delete_at(index_or_header)
         else                 # by header
+          if index_or_header.is_a? Integer
+            @headers.delete_at(index_or_header)
+          else
+            @headers.delete(index_or_header)
+          end
```
kou: Can you add a test for this case?

lib/csv/table.rb
```diff
@@ -285,7 +291,8 @@ def delete_if(&block)
         @table.delete_if(&block)
       else                                      # by header
         deleted = []
-        headers.each do |header|
+        initial_headers = headers.dup
```
kou: Is `dup` needed?

shibasaka: I've fixed it(because `dup` had been duplicated in `headers` method).

lib/csv/table.rb
shibasaka: I've added tests of deleting headers by index and header.

lib/csv/table.rb
```diff
@@ -23,7 +23,8 @@ class Table
     # * length()
     # * size()
     #
-    def initialize(array_of_rows)
+    def initialize(array_of_rows, headers=nil)
+      @headers = headers || []
```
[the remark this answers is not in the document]
shibasaka: I've changed it and restored argument in test code.

lib/csv/table.rb
```diff
@@ -23,8 +23,19 @@ class Table
...
+      if headers.nil?
+        @table = array_of_rows
+        if @table.empty?
+          @headers = []
+        else
+          @headers = @table.first.headers
+        end
+      else
+        @headers = headers
+        @table = array_of_rows
```
esparta: There's a duplication of code here: the `@table` assignment. We should keep only one at the beginning of `initialize`.

lib/csv/table.rb
```diff
@@ -285,7 +301,8 @@ def delete_if(&block)
         @table.delete_if(&block)
       else                                      # by header
         deleted = []
-        headers.each do |header|
+        initial_headers = headers
```
esparta: I can see you want to iterate through the initial headers of the table, but at this line `headers` is equals `@headers.dup`. Looks like we don't need this extra assignment.

lib/csv/table.rb
```diff
@@ -23,8 +23,19 @@ class Table
     # * length()
     # * size()
     #
-    def initialize(array_of_rows)
-      @table = array_of_rows
+    def initialize(array_of_rows, headers=nil)
+      if headers.nil?
```
esparta: This is only style, and also opinion: can we invert this to use ruby's truthy?
```ruby
if headers
  @headers = headers
else
  if @table.empty?
   # ...
  end
end
```

lib/csv/table.rb
```diff
@@ -23,8 +23,19 @@ class Table
     # * length()
     # * size()
     #
-    def initialize(array_of_rows)
-      @table = array_of_rows
+    def initialize(array_of_rows, headers=nil)
```
esparta: Since is a public class, can you add a documentation about the new parameter?

shibasaka: I've added the documentation about headers argument.

lib/csv/table.rb
shibasaka: That it is clear, I've fixed it.

lib/csv/table.rb
shibasaka: You're right, I've fixed it.

lib/csv/table.rb
shibasaka: I've moved the @table assignment to the beginning of initialize.

kou:
Thanks.
I've merged.

I've changed `headers` to keyword argument from optional argument and added a missing test to parse header only CSV.

shibasaka:
Thanks!

PR #48, opened 2018-10-28
Removing branch on raw_encoding

esparta:
Turns out that `StringIO` objects responds to both `:internal_encoding` & `:external_encoding` so it doesn't execute that condition and the determinationof the encoding


kou:
Good catch!

PR #54, opened 2018-12-17
Fix some typos from my previous PR

zverok:
Some dumb typos were introduced in my prev. PR: https://github.com/ruby/csv/pull/32
Just fixing them before Xmas.


kou:
Good catch!

Issue #62, opened 2018-12-26
CSV.generate returns unexpected encoding string


Watson1978:
Since Ruby's internal encoding is UTF-8, I expect `CSV.generate` returns `UTF-8` string as well
However, `CSV.generate` returns `ASCII-8BIT` string by default.

https://github.com/ruby/csv/blob/9b81ece22d541c9083f308e1cb4f71a49b75a4b3/lib/csv.rb#L537

Seems that `String.new` returns `ASCII-8BIT` string.

I want to know whether this is right behavior in Ruby 2.6 's CSV library. Thank you!!!

## Ruby 2.6

```
$ ruby -v t.rb
ruby 2.6.0p0 (2018-12-25 revision 66547) [x86_64-darwin18]
#<Encoding:ASCII-8BIT>
"\xE5\xBA\x97\xE5\x90\x8D,\xE5\x90\x8D\xE7\xA7\xB0,\xE7\xA8\x8E\xE9\xA1\x8D\xE5\x8C\xBA\xE5\x88\x86,\xE6\x95\xB0\xE9\x87\x8F,\xE7\xA8\x8E\xE9\xA1\x8D\n"
```

## Ruby 2.5

```
$ ruby -v t.rb
ruby 2.5.3p105 (2018-10-18 revision 65156) [x86_64-darwin18]
#<Encoding:UTF-8>
"店名,名称,税額区分,数量,税額\n"
```

## Test code

```
require 'csv'
csv_string = CSV.generate do |csv|
  csv << ["店名", "名称", "税額区分", "数量", "税額"]
end
p csv_string.encoding
p csv_string
```

kou:
Ah, it's not expected.
This is caused by 25186cff11d01179f6be6685c30d222f1ae2b9f8 for `frozen_string_literal`.

We should use `+""` instead of `String.new`. If we use `+""`, we can choose the encoding of the first row.

Can you send a pull request with a test?

PR #63, opened 2018-12-26
Fix string encoding of CSV.generate returns

Watson1978:
Fix https://github.com/ruby/csv/issues/62

This PR will change default string encoding to UTF-8 in `CSV.generate()`.


kou:
Thanks.

TODO:

  * Extract encoding test to other test method.
  * Add a test for not UTF-8 encoding case such as EUC-JP.
  * Do related work for `generate_line`.

PR #65, opened 2018-12-26
Fix parse case of empty string with headers true

### LINKED ISSUE #64: `CSV.parse('', headers: true)` was broken?
If `headers: true` option was given, `CSV.parse()` returns `CSV::Table` object even if empty string will be passed as argument with Ruby 2.5.0.

However, in Ruby 2.6, the above behavior was incomatible. 

I want to know whether it is expected. Thank you!!

### Ruby 2.6

```
$ ruby -v ttt.rb
ruby 2.6.0p0 (2018-12-25 revision 66547) [x86_64-darwin18]
[]
Traceback (most recent call last):
ttt.rb:4:in `<main>': undefined method `headers' for []:Array (NoMethodError)
```

### Ruby 2.5

```
$ ruby -v ttt.rb
ruby 2.5.3p105 (2018-10-18 revision 65156) [x86_64-darwin18]
#<CSV::Table mode:col_or_row row_count:1>
[]
```

### Test code
```ruby
require 'csv'
row = CSV.parse('', headers: true)
p row
p row.headers
```

284km:
Fix #64

In this case "undefined method `headers' for []:Array (NoMethodError)"
was occurring.

```
row = CSV.parse('', headers: true)
p row
p row.headers
```

when CSV.parse with (headers: true) option,
return CSV::Table class is expected, but it was returned Array as below.

```
def read
  rows = to_a #=> []
  headers = parser.headers #=> nil
  if headers
    Table.new(rows, headers: headers)
  else
    rows
  end
end
```

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1179,7 +1179,7 @@ class CSV
   def read
     rows = to_a
     headers = parser.headers
-    if headers || rows.empty?
+    if headers || parser.use_headers
       Table.new(rows, headers: headers)
     else
       rows
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1179,7 +1179,7 @@ class CSV
   def read
     rows = to_a
     headers = parser.headers
-    if headers
+    if headers || rows.empty?
       Table.new(rows, headers: headers)
     else
       rows
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1178,9 +1178,8 @@ class CSV
   #
   def read
     rows = to_a
-    headers = parser.headers
-    if headers || parser.use_headers
-      Table.new(rows, headers: headers)
+    if parser.use_headers?
+      Table.new(rows, headers: parser.headers)
     else
       rows
     end
--- a/lib/csv/parser.rb
+++ b/lib/csv/parser.rb
@@ -289,7 +289,7 @@ class CSV
       end
     end
 
-    def use_headers
+    def use_headers?
       @use_headers
     end
 
--- a/lib/csv/parser.rb
+++ b/lib/csv/parser.rb
@@ -289,6 +289,10 @@ class CSV
       end
     end
 
+    def use_headers
+      @use_headers
+    end
+
     private
     def prepare
       prepare_variable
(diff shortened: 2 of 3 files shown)


kou:
We shouldn't return `Table` when `headers:` isn't specified.

284km:
I understand. Changed to judge by CSV::Parser#use_headers.

lib/csv.rb
```diff
@@ -1179,7 +1179,7 @@ def each(&block)
   def read
     rows = to_a
     headers = parser.headers
-    if headers
+    if headers || parser.use_headers
```
kou: Can we remove `headers ||`?

test/csv/test_headers.rb
```diff
@@ -315,4 +315,9 @@ def test_nil_row_header
     assert_equal([["A"], [nil]],
                  [row.headers, row.fields])
   end
+
+  def test_empty_string_value_with_headers_true
+    row = CSV.parse('', headers: true)
```
kou: It must be `table`.

test/csv/test_headers.rb
```diff
@@ -315,4 +315,9 @@ def test_nil_row_header
...
+    assert_equal([], row.headers)
```
kou: `assert_equal(CSV::Table.new([], []), table)` is better because it shows more information (such as class ans instance variables) on failure.

FYI: https://www.clear-code.com/blog/2011/2/28.html

lib/csv/parser.rb
```diff
@@ -289,6 +289,10 @@ def parse(&block)
       end
     end
 
+    def use_headers
```
kou: `use_headers?` is better.

lib/csv.rb
284km: Yes. Only parser.use_headers? is enough condition.

test/csv/test_headers.rb
284km: Fixed it.

test/csv/test_headers.rb
284km: Thank you for information. Fixed it.

lib/csv/parser.rb
284km: Definedly. Fixed it.

Issue #68, opened 2019-01-06
CSV.each rewinding and including header row


maschwenk:
#### Reproduction script
```ruby
require 'csv'

data = CSV.new(<<~ROWS, headers: true)
  Name,Department,Salary
  Bob,Engineering,1000
ROWS

data.each do |row|
  puts row.to_s
end

puts 'second loop'

data.each do |row|
  puts row.to_s
end
```

#### ruby 2.6.0p0 (2018-12-25 revision 66547) [x86_64-darwin18]
`ruby csv_bug.rb`
```
Bob,Engineering,1000
second loop
Name,Department,Salary  <--- 🔥this should not be here 🔥 
Bob,Engineering,1000
```

#### ruby 2.5.3p105 (2018-10-18 revision 65156) [x86_64-darwin18]
`ruby csv_bug.rb`
```
Bob,Engineering,1000
second loop
```
#### ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin17]
`ruby csv_bug.rb`
```
Bob,Engineering,1000
second loop
```
____

#### What I think is going on
I may take a crack at fixing myself, but almost _certain_ this isn't desired behavior. 

It seems the row pointer is being rewound when calling `.each` twice, but on the second iteration the rewind is including the header line -- even though I've specified `headers: true`

Side note, in the future I think `.each` should always rewind the row counter after execution, which appears what it's trying to do in `2.6.0` but it's accidentally including the header row
____

#### Side note
Bob should ask for a raise

kou:
Thanks for your report.
I've fixed this. It means that the second `#each` doesn't yield anything.

Auto `#rewind` in each `#each` isn't good idea. Because `CSV` provides IO like API. IO doesn't rewind automatically. `CSV` should not rewind automatically too.

PR #69, opened 2019-01-07
`CSV::Row#include?` aliased twice

maschwenk:
In working on #68 I noticed `include?` is aliased twice on `CSV::Row` (with the latter alias taking precedence due to order). 

The existing `header?` method looked simplest so I got rid of `has_key?` and aliased them all to `header?`.

Feels like having duplicate aliases defined is destined to create a bug later so figured I'd fix it.

### DIFF
--- a/lib/csv/row.rb
+++ b/lib/csv/row.rb
@@ -123,6 +123,15 @@ class CSV
       end
     end
 
+    # Returns +true+ if there is a field with the given +header+.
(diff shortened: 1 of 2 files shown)


maschwenk:
@kou I've addressed your recommendations in 4b8f073. Let me know if you need me to squash/rebase/etc.

kou:
Oh... Not squashed...

PR #70, opened 2019-01-07
[CHORE] Duplicate definition of _header

maschwenk:
I'm unsure what this code was originally trying to work around but I think this change makes more sense and is functionally the same.

Perhaps at some point calculating `headers` was expensive?

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1244,7 +1244,6 @@ class CSV
       end
     end
     _headers = headers
-    _headers = headers
     str << " headers:" << _headers.inspect if _headers
     str << ">"
     begin


lib/csv.rb
```diff
@@ -1243,9 +1243,11 @@ def inspect
         str << " " << attr_name << ":" << a.inspect
       end
     end
-    _headers = headers
-    _headers = headers
```
kou: I prefer to just remove this line.
I don't want to call the same method twice.

maschwenk:
@kou I've addressed your recommendations in 42893fa. Let me know if you need me to squash/rebase/etc.

PR #73, opened 2019-01-23
Add ability to convert a row on writing

### LINKED ISSUE #46: Converters doesn't work on generate_line
When you call

```ruby
CSV.generate_line(['a'], converters: lambda { |field| field.prepend '=' } )
``` 

doesn't do anything.

Checking the code, and correct me if I'm wrong, the only supported options when using `generate_line` are https://github.com/ruby/csv/blob/9ccdfe368c4446f8925daeca7d12d76d9855e445/lib/csv.rb#L1083

The "problem" in my opinion is that the documentation says that it can be used any option `new` understands: https://github.com/ruby/csv/blob/9ccdfe368c4446f8925daeca7d12d76d9855e445/lib/csv.rb#L563-L567

I can try to make changes to make it work if you guys point me what should be the expected behaviour.

paniko0:
Just adding the ability to use `write_converters` when creating CSV data. It is supposed to fix #46. I am allowing the builtins converters to be used, but I'm not sure if you guys want to keep this behaviour.

Another thing I'm not entirely sure is what else I can test, so it would be nice if someone can suggest more sensitive points in the code that can be covered.

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -910,6 +910,7 @@ class CSV
                  external_encoding: nil,
                  encoding: nil,
                  nil_value: nil,
+                 write_converters: nil,
                  empty_value: "",
                  quote_empty: true)
     raise ArgumentError.new("Cannot parse nil as CSV") if data.nil?
@@ -924,6 +925,7 @@ class CSV
     }
     @initial_converters = converters
     @initial_header_converters = header_converters
+    @initial_write_converters = write_converters
 
     @parser_options = {
       column_separator: col_sep,
@@ -952,6 +954,7 @@ class CSV
       row_separator: row_sep,
       quote_character: quote_char,
       quote_empty: quote_empty,
+      write_converters: write_converters,
     }
 
     @writer = nil
@@ -1325,11 +1328,7 @@ class CSV
       builtin_converters: Converters,
     }
     options = @base_fields_converter_options.merge(specific_options)
-    fields_converter = FieldsConverter.new(options)
-    normalize_converters(@initial_converters).each do |name, converter|
-      fields_converter.add_converter(name, &converter)
-    end
-    fields_converter
+    parse_converters(@initial_converters, options)
   end
 
   def header_fields_converter
@@ -1342,8 +1341,25 @@ class CSV
       accept_nil: true,
     }
     options = @base_fields_converter_options.merge(specific_options)
+    parse_converters(@initial_header_converters, options)
+  end
+
+  def writer_fields_converter
+    @writer_fields_converter ||= build_writer_fields_converter
+  end
+
+  def build_writer_fields_converter
+    specific_options = {
+      builtin_converters: Converters
+    }
+    options = @base_fields_converter_options.merge(specific_options)
+
+    parse_converters(@initial_write_converters, options)
+  end
+
+  def parse_converters(initial_converters, options)
     fields_converter = FieldsConverter.new(options)
-    normalize_converters(@initial_header_converters).each do |name, converter|
+    normalize_converters(initial_converters).each do |name, converter|
       fields_converter.add_converter(name, &converter)
     end
     fields_converter
@@ -1363,7 +1379,8 @@ class CSV
   end
 
   def writer_options
-    @writer_options.merge(header_fields_converter: header_fields_converter)
+    @writer_options.merge(header_fields_converter: header_fields_converter,
+                          writer_fields_converters: writer_fields_converter)
   end
 end
 
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1349,7 +1349,7 @@ class CSV
   end
 
   def build_writer_fields_converter
-    generate_fields_converter(@initial_write_converters, {})
+    generate_fields_converter(@initial_write_converters, @base_fields_converter_options)
   end
 
   def generate_fields_converter(initial_converters, options)
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1328,7 +1328,7 @@ class CSV
       builtin_converters: Converters,
     }
     options = @base_fields_converter_options.merge(specific_options)
-    build_converters(@initial_converters, options)
+    generate_fields_converter(@initial_converters, options)
   end
 
   def header_fields_converter
@@ -1341,7 +1341,7 @@ class CSV
       accept_nil: true,
     }
     options = @base_fields_converter_options.merge(specific_options)
-    build_converters(@initial_header_converters, options)
+    generate_fields_converter(@initial_header_converters, options)
   end
 
   def writer_fields_converter
@@ -1349,14 +1349,10 @@ class CSV
   end
 
   def build_writer_fields_converter
-    specific_options = {
-      builtin_converters: Converters,
-    }
-    options = @base_fields_converter_options.merge(specific_options)
-    build_converters(@initial_write_converters, options)
+    generate_fields_converter(@initial_write_converters, {})
   end
 
-  def build_converters(initial_converters, options)
+  def generate_fields_converter(initial_converters, options)
     fields_converter = FieldsConverter.new(options)
     normalize_converters(initial_converters).each do |name, converter|
       fields_converter.add_converter(name, &converter)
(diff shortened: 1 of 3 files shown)


lib/csv.rb
```diff
@@ -952,6 +954,7 @@ def initialize(data,
       row_separator: row_sep,
       quote_character: quote_char,
       quote_empty: quote_empty,
+      write_converters: write_converters,
```
kou: Could you use `converters`? `write_` is redundant here. Because this is `@writer_options`.

lib/csv.rb
```diff
@@ -1363,7 +1379,8 @@ def writer
   end
 
   def writer_options
-    @writer_options.merge(header_fields_converter: header_fields_converter)
+    @writer_options.merge(header_fields_converter: header_fields_converter,
+                          writer_fields_converters: writer_fields_converter)
```
kou: Could you use `fields_converter`? `writer_` is redundant for writer. It's not plural.

lib/csv.rb
```diff
@@ -1325,11 +1328,7 @@ def build_fields_converter
       builtin_converters: Converters,
     }
     options = @base_fields_converter_options.merge(specific_options)
-    fields_converter = FieldsConverter.new(options)
-    normalize_converters(@initial_converters).each do |name, converter|
-      fields_converter.add_converter(name, &converter)
-    end
-    fields_converter
+    parse_converters(@initial_converters, options)
```
kou: `parse_` is strange for me. Because it doesn't parse converters.
How about `build_fields_converter`?

lib/csv.rb
```diff
@@ -1342,8 +1341,25 @@ def build_header_fields_converter
...
+    parse_converters(@initial_header_converters, options)
+  end
+
+  def writer_fields_converter
+    @writer_fields_converter ||= build_writer_fields_converter
+  end
+
+  def build_writer_fields_converter
+    specific_options = {
+      builtin_converters: Converters
```
kou: It's strange. `Converters` converts `String` to `Integer` or something.
We need to convert anything to `String` on write.

lib/csv/writer.rb
```diff
@@ -31,6 +31,8 @@ def <<(row)
       @headers ||= row if @use_headers
       @lineno += 1
 
+      apply_converters_to(row, lineno)
```
kou: We need to receive converted result such as `row = @converter.convert(row, ...) if @converter`.

lib/csv/writer.rb
```diff
@@ -149,5 +151,10 @@ def quote(field)
         end
       end
     end
+
+    def apply_converters_to(row, lineno)
+      converter = @options[:writer_fields_converters]
```
kou: Could you avoid fetching from `@options` each time? We need high performance.
Instance variable is faster than `Hash#[]`.

test/csv/write/test_general.rb
```diff
@@ -205,6 +205,12 @@ def test_encoding_euc_jp
     assert_equal(%Q[あ,い,う#{$INPUT_RECORD_SEPARATOR}].encode("EUC-JP"),
                  generate_line(row))
   end
+
+  def test_write_converters
+    assert_equal(%Q[=a,=b,=c\n],
+                 generate_line(["a", "b", "c"],
+                               write_converters: ->(value) { value.prepend('=') } ))
```
kou: Could you avoid using destructive method in converter`?
Converter should return converted value.

lib/csv.rb
paniko0: I was a bit confused about this one. So I completely removed the builtin_converters

test/csv/write/test_general.rb
paniko0: Another one I was a bit confused. I think I fixed it but it would be nice if you can take a look at it again.

kou: `value.prepend('=')` is a destructive method call.
`'=' + value` isn't a destructive method call.

```ruby
a = "x"
"=" + a
p(a) # => "x"
a.prepend("=")
p(a) # => "=x"
```

lib/csv/writer.rb
```diff
@@ -149,5 +152,9 @@ def quote(field)
         end
       end
     end
+
+    def apply_converters_to(row, lineno)
+      @converters.convert(row, nil, lineno)
+    end
```
kou: Could you remove this method and use `@converters.convert` directly?
It looks needless abstraction to me.

lib/csv.rb
```diff
@@ -1342,8 +1341,24 @@ def build_header_fields_converter
...
+    build_converters(@initial_header_converters, options)
+  end
+
+  def writer_fields_converter
+    @writer_fields_converter ||= build_writer_fields_converter
+  end
+
+  def build_writer_fields_converter
+    specific_options = {
+      builtin_converters: Converters,
```
kou: This still exists.

lib/csv.rb
```diff
@@ -1342,8 +1341,24 @@ def build_header_fields_converter
...
+    }
+    options = @base_fields_converter_options.merge(specific_options)
```
kou: `@base_fields_converter_options` isn't needed. It's for options on parsing.

lib/csv.rb
```diff
@@ -1342,8 +1341,24 @@ def build_header_fields_converter
...
+    build_converters(@initial_write_converters, options)
+  end
+
+  def build_converters(initial_converters, options)
```
kou: This builds a fields converter from converters. This doesn't build converters.
Do you have more good name?

paniko0: What do you think about `merge_` since it merges the defaults and any custom converter or maybe just call it `fields_converter` since it returns an instance of `FieldsConverter`?

lib/csv/writer.rb
paniko0: I added this abstraction because I thought it would increase readability, but I agree it's a little bit overkill in this scenario.

lib/csv.rb
paniko0: I am so sorry. I changed it in my last change but I didn't commit it.

lib/csv.rb
paniko0: I the end I named it `generate_fields_converter` since it is exactly what this method does. Let me know if you think this is a good name :+1:

lib/csv/writer.rb
```diff
@@ -18,6 +18,7 @@ def initialize(output, options)
       if @options[:write_headers] and @headers
         self << @headers
       end
+      @converters = @options[:fields_converter] unless @options[:fields_converter].empty?
```
kou: This is a converter not converter"s".
`unless @options[:fields_converter].empty?` is needless.

lib/csv.rb
kou: It's better than before. But `build_` and `generate_` may be confused. "generate_" is OK in this pull request. This is a separated issue. I'll consider better name later.

lib/csv/writer.rb
paniko0: Hey @kou, thanks for the feedback!

I named it `converters` because this instance (FieldsConverter) can hold zero, one or many converters. Do you think it's better if I just name it `@converter` or should the variable be named as `fields_converter`? I think the second option explicit tells any other developer what instance they are handling.

Regarding the `.empty?` call, if I don't add it, we will call `@converters.convert`  regardless of any converter informed by the user. It is not a call to `Hash#empty?` but it is a call to `FieldsConverter#empty?`.

kou: `@fields_converter` is better because we use `@fields_converter` in `CSV::Parser`.

OK. If we can confirm the `empty?` check optimization has performance improvement by benchmark, we can accept it.

NOTE: We may accept more new converters after we instantiate writer object. `CSV#convert` is a method that adds a new converter on parsing after we instantiate parser object. We may add `CSV#write_convert` method in the future. We need to remove the `empty?` check optimization when we add the method.

paniko0: @kou finished the suggestions.

I have changed the variable name and removed the `.empty?` check. I also had to add the base converters so nil values are converted to empty string.

kou: OK. But we should use different names for them such as `write_empty_value` and `write_nil_value`. Because context is different on parsing and writing.
I'll do this after I merge this.

PR #76, opened 2019-02-04
Fix CSV#shift regression (introduced in 3.0.2)

alyssais:
Prior to 39da059c6d363136a80867a6cdb7f2ac42e48ecc, shifting a value from a `CSV`, and then using an `Enumerable` method, would not include the shifted value in the result of the `Enumerable` method.

This behavior made sense. The `#shift` methods on `Array` and `Hash` both behave this same way, because they both remove an element from the underlying data structure. It makes sense for `CSV#shift` to work like this to, as it used to.

To fix this regression, there needs to be an internal position that is advanced when calling `#shift`, which `#each` will then start from. By re-implementing `#each` in terms of `Enumerator#next`, we get this for free. This also allows `#shift` to be re-implemented in terms of `#each`.

It might seem that this would break using `#each` multiple times without sending `#rewind` in between, but it turns out this has never been supported anyway. I've added a test for this behaviour so as to document it.

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1173,7 +1173,7 @@ class CSV
     return to_enum(__method__) unless block_given?
     @parser_enumerator ||= parser.parse
     begin
-      loop do
+      while true
         yield @parser_enumerator.next
       end
     rescue StopIteration
@@ -1210,8 +1210,9 @@ class CSV
   # The data source must be open for reading.
   #
   def shift
+    @parser_enumerator ||= parser.parse
     begin
-      each.next
+      @parser_enumerator.next
     rescue StopIteration
       nil
     end
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1169,8 +1169,16 @@ class CSV
   #
   # The data source must be open for reading.
   #
-  def each(&block)
-    parser.parse(&block)
+  def each
+    return to_enum(__method__) unless block_given?
+    @parser_enumerator ||= parser.parse
+    begin
+      loop do
+        yield @parser_enumerator.next
+      end
+    rescue StopIteration
+    end
+    self
   end
 
   #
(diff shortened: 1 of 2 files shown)


lib/csv.rb
```diff
@@ -1202,9 +1210,8 @@ def header_row?
   # The data source must be open for reading.
   #
   def shift
-    @parser_enumerator ||= parser.parse
     begin
-      @parser_enumerator.next
+      each.next
```
kou: This change causes performance regression with this benchmark: https://github.com/ruby/csv/blob/master/benchmark/shift.yaml

The following result is generated by changed YAML file to just show one case:

```console
% rake benchmark:shift:small
RUBYLIB= BUNDLER_ORIG_RUBYLIB= N_COLUMNS=10 /tmp/local/bin/ruby -v -S benchmark-driver /home/kou/work/ruby/csv/benchmark/shift.yaml
ruby 2.7.0dev (2019-02-02 trunk 66993) [x86_64-linux]
Warming up --------------------------------------
            unquoted    190.213 i/s -     200.000 times in 1.051452s (5.26ms/i)
Calculating -------------------------------------
                         master  shift-each 
            unquoted    187.657      76.664 i/s -     570.000 times in 3.037459s 7.435047s

Comparison:
                         unquoted
              master:       187.7 i/s 
          shift-each:        76.7 i/s - 2.45x  slower
```

We can resolve the regression by reverting this change.

lib/csv.rb
```diff
@@ -1169,8 +1169,16 @@ def header_convert(name = nil, &converter)
   #
   # The data source must be open for reading.
   #
-  def each(&block)
-    parser.parse(&block)
+  def each
+    return to_enum(__method__) unless block_given?
+    @parser_enumerator ||= parser.parse
+    begin
+      loop do
```
kou: Could you use `while true`? It's a bit faster than `loop`.

PR #78, opened 2019-02-20
Fix line number in invalid byte sequence error

alyssais:
This didn't take into account that the sample could be composed of multiple lines, and so the index of the malformed line in the sample wasn't taken into account when calculating the line number.

I chose to just change the existing test for this to have the error on line two, rather than writing a whole new test, because I think it's much more likely that 1 will be returned incorrectly than 2, so this way it's just an all around better test.

### DIFF
--- a/lib/csv/parser.rb
+++ b/lib/csv/parser.rb
@@ -575,9 +575,9 @@ class CSV
           string = @samples[0]
         end
         if string
-          unless string.valid_encoding?
+          if index = string.lines(@row_separator).index { |l| !l.valid_encoding? }
             message = "Invalid byte sequence in #{@encoding}"
-            raise MalformedCSVError.new(message, @lineno + 1)
+            raise MalformedCSVError.new(message, @lineno + index + 1)
           end
           Scanner.new(string)
         else
--- a/test/csv/test_encodings.rb
+++ b/test/csv/test_encodings.rb
@@ -256,12 +256,13 @@ class TestCSVEncodings < Test::Unit::TestCase
   end
 
   def test_invalid_encoding_row_error
-    csv = CSV.new("invalid,\xF8\r\nvalid,x\r\n".force_encoding("UTF-8"),
-                  encoding: "UTF-8")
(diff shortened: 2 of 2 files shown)


kou:
Did you confirm performance impact by this change?

alyssais:
I ran the "parse" benchmark before and after, and there was not really any difference. With this change, it actually appeared slightly faster, although I'm guessing that's just noise.

test/csv/test_encodings.rb
```diff
@@ -256,12 +256,11 @@ def test_row_separator_detection_with_invalid_encoding
   end
 
   def test_invalid_encoding_row_error
-    csv = CSV.new("invalid,\xF8\r\nvalid,x\r\n".force_encoding("UTF-8"),
-                  encoding: "UTF-8")
     error = assert_raise(CSV::MalformedCSVError) do
-      csv.shift
+      CSV.parse("valid,x\r\ninvalid,\xF8\r\n".force_encoding("UTF-8"),
+                encoding: "UTF-8")
     end
```
kou: We should not change to use `CSV.parse`. It accepts a case that `CSV.new` raises an error.
We should use `csv.shift` twice in the `assert_raise` block.

lib/csv/parser.rb
```diff
@@ -575,9 +575,9 @@ def build_scanner
           string = @samples[0]
         end
         if string
-          unless string.valid_encoding?
+          if index = string.lines.index { |l| !l.valid_encoding? }
```
kou: We need to respect `@row_separator`: `string.lines(@raw_separator)`.

alyssais: I assume that `@raw_separator` was just a typo for `@row_separator`?

alyssais:
Thank you for the review. I've addressed both comments.

lib/csv/parser.rb
kou: Ah, sorry. You're right.

kou:
Thanks.
I've merged.

PR #80, opened 2019-03-13
Fix CSV delegation to missing StringIO methods

gsamokovarov:
If you create a `CSV` from raw content like:

    csv = CSV.new("h1,h2")

You'll get a `NoMethodError` when calling `csv.path` but still get true when
you call `csv.respond_to?(:path)`.

This tricks 3rd party libraries like `carrierwave` which try to call path
on their input, if it responds to it.

See https://github.com/carrierwaveuploader/carrierwave/blob/a91ab69fdd8052cdf5a5e48ef8baf40939e441fb/lib/carrierwave/sanitized_file.rb#L109-L123

This stops me from passing `CSV` objects as `StringIO`'s into `carrierwave`
uploads, for example, but the problem can also be manifested in other
3rd party libraries, as responding to a method and returning a
NoMethodError when calling it is still an unexpected behavior.

I have gone through the `CSV` delegation scheme and made sure that every
method that `StringIO` doesn't respond to returns a meaningful zero value
and does not raise a `NoMethodError` while used through a `CSV` Instance.

Ref. https://github.com/ruby/ruby/pull/2094

### DIFF
--- a/test/csv/interface/test_delegation.rb
+++ b/test/csv/interface/test_delegation.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: false
+
+require_relative "../helper"
+
+class TestCSVInterfaceDelegation < Test::Unit::TestCase
+  def test_stringio_missing_methods_delegation
+    csv = CSV.new("h1,h2")
+
+    assert_raise(NotImplementedError) { csv.flock(0) }
+    assert_raise(NotImplementedError) { csv.ioctl(0) }
+    assert_raise(NotImplementedError) { csv.stat }
+    assert_raise(NotImplementedError) { csv.to_i }
+
+    assert_equal(false, csv.binmode?)
+    assert_equal(nil, csv.path)
+    assert_instance_of(StringIO, csv.to_io)
+  end
+end
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1106,12 +1106,44 @@ class CSV
   ### IO and StringIO Delegation ###
 
   extend Forwardable
-  def_delegators :@io, :binmode, :binmode?, :close, :close_read, :close_write,
+  def_delegators :@io, :binmode, :close, :close_read, :close_write,
                        :closed?, :external_encoding, :fcntl,
-                       :fileno, :flock, :flush, :fsync, :internal_encoding,
-                       :ioctl, :isatty, :path, :pid, :pos, :pos=, :reopen,
-                       :seek, :stat, :string, :sync, :sync=, :tell, :to_i,
-                       :to_io, :truncate, :tty?
+                       :fileno, :flush, :fsync, :internal_encoding,
+                       :isatty, :pid, :pos, :pos=, :reopen,
+                       :seek, :string, :sync, :sync=, :tell,
+                       :truncate, :tty?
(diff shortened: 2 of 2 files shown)


test/csv/interface/test_delegation.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: false
+
+require_relative "../helper"
+
+class TestCSVInterfaceRead < Test::Unit::TestCase
```
kou: Could you use `Delegation` instead of `Read`?

test/csv/interface/test_delegation.rb
```diff
@@ -0,0 +1,17 @@
...
+  def test_stringio_missing_methods_delegation
+    csv = CSV.new("h1,h2")
+
+    assert_equal(nil, csv.binmode?)
```
kou: Could you use `false` for the expected value?

test/csv/interface/test_delegation.rb
```diff
@@ -0,0 +1,17 @@
...
+    assert_equal(nil, csv.flock(0))
```
kou: I think that `NotImplementedError` is suitable instead of returning `nil`.
Because an application assumes this object is locked by calling this method is dangerous. This method never acquire a lock.

test/csv/interface/test_delegation.rb
```diff
@@ -0,0 +1,17 @@
...
+    assert_equal(nil, csv.ioctl(0, 0))
```
kou: I think that `NotImplementedError` is suitable instead of returning `nil`.
Because `IO#ioctl` raises `NotImplementedError` on environment that doesn't have `ioctl(2)` such as Windows.

test/csv/interface/test_delegation.rb
```diff
@@ -0,0 +1,17 @@
...
+    assert_equal(nil, csv.path)
+    assert_equal(nil, csv.stat)
+    assert_equal(0, csv.to_i)
```
kou: We should not return `0`. Because `0` is a valid file descriptor. Normally, it's used for the standard input.
I think that `NoMethodError` is better.

gsamokovarov:
Thanks for taking a look, @kou! I have changed some of the operations to return `NotImplementedError`s.

test/csv/interface/test_delegation.rb
gsamokovarov: Good catch! Got the system calls to raise `NotImplementedError` in `StringIO` and the likes.

test/csv/interface/test_delegation.rb
gsamokovarov: Good catch! 🙇‍♂️

test/csv/interface/test_delegation.rb
gsamokovarov: Thanks for the catch!

kou:
Did you push? It seems that there is no new commit.

gsamokovarov:
Yes! I did forget... 🤦‍♂️ Can you check now? 😅

gsamokovarov:
Thank you!

PR #90, opened 2019-04-20
Add options "undef", "replace" form CSV#open

yuuna:
I want to use CSV#open with undef: :repace, replace: "*" options form wrong character.

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -907,6 +907,13 @@ class CSV
   #                                       +true+, CSV will strip string. The
   #                                       length of string must be 1.
   #
+  # <b><tt>:undef</tt></b>::              When Converting encoding with 
+  #                                       "undef: :replace", you will replace
+  #                                       character to replace.
+  #                                       
+  # <b><tt>:replace</tt></b>::            Replace undef character to this value
+  #                                       
+  #
   # See CSV::DEFAULT_OPTIONS for the default settings.
   #
   # Options cannot be overridden in the instance methods for performance reasons,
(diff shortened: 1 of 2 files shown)


lib/csv.rb
```diff
@@ -936,7 +943,9 @@ def initialize(data,
                  write_converters: nil,
                  write_nil_value: nil,
                  write_empty_value: "",
-                 strip: false)
+                 strip: false,
+                 undef: nil,
+                 replace: nil)
```
kou: Should we need to accept `:undef` and `:replace` in `CVS#initialize`?
It seems that we just need to accept them in `CSV.open`. Then we can remove them from `options` in `CSV.open` arguments.

test/csv/interface/test_read.rb
```diff
@@ -264,6 +264,14 @@ def test_new_nil
     end
   end
 
+  def test_foreach_undef_replace
+    rows = []
+    CSV.foreach(@input.path, col_sep: "\t", row_sep: "\r\n", undef: :replace, replace: "*").each do |row|
+      rows << row
+    end
+    assert_equal(@rows, rows)
+  end
```
kou: We need to new test data that fails without these new options.

lib/csv.rb
yuuna: I will fix  the arguments in CSV.open.

test/csv/interface/test_read.rb
yuuna: I will make a new test data.

kou:
Orphan.

PR #95, opened 2019-06-07
Fix typo in docs

y-yagi:
This fix that the execution result of the next line is different from the actual result.


kou:
Good catch!

PR #96, opened 2019-07-08
Suppress warnings

nobu:
Suppress warnings in  verbose mode.

```
test/csv/interface/test_read_write.rb:9: warning: assigned but unused variable - rows
csv/lib/csv/writer.rb:35: warning: instance variable @fields_converter not initialized
```

### DIFF
--- a/lib/csv/writer.rb
+++ b/lib/csv/writer.rb
@@ -14,6 +14,7 @@ class CSV
       @output = output
       @options = options
       @lineno = 0
+      @fields_converter = nil
       prepare
       if @options[:write_headers] and @headers
         self << @headers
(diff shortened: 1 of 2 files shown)


lib/csv/writer.rb
```diff
@@ -14,6 +14,7 @@ def initialize(output, options)
       @output = output
       @options = options
       @lineno = 0
+      @fields_converter = nil
```
kou: Ah, I need to confirm why this is needed. (TODO)

nobu: In `CSV::Writer#initialize`:

```ruby
      if @options[:write_headers] and @headers
        self << @headers
      end
```

and `CSV::Writer#<<` refers `@fields_converter`.

```ruby
      row = @fields_converter.convert(row, nil, lineno) if @fields_converter
```

nobu: I wasn't sure if the converter can be applied to headers too.

kou: Thanks. It should not be applied to headers.
So this is OK.

PR #98, opened 2019-08-13
Doesn't work seeking position when set input class of StringIO

katsyoshi:
Doesn't work seeking position when set input class of StringIO.

I add position for StringIOs.

example: https://gist.github.com/katsyoshi/bfd86639d9464f60f27b26e20a85c720

### DIFF
--- a/lib/csv/parser.rb
+++ b/lib/csv/parser.rb
@@ -192,7 +192,7 @@ class CSV
         input = @inputs.first
         case input
         when StringIO
-          string = input.string[input.pos..-1]
+          string = input.read
           raise InvalidEncoding unless string.valid_encoding?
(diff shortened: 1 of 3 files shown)


esparta:
I'm seeing the failure on your gist, can you add a test case for this `StringIO` input?

katsyoshi:
@kou @esparta thank you comments.
 I replaced string range chars to read. I added test using StringIO.

kou:
Thanks. I've fixed some remaining problems:

  * We don't need to use `rewind` when we use `seek`.
  * There are some needless `rewind`s.
  * The added test doesn't use `StringIO`. We need to use `StringIO` for this case.

PR #100, opened 2019-08-16
Fix typos in CSV::Row docs

vbrazo:
- Fix typos in `CSV::Row` documentation
- Remove unnecessary spaces

### DIFF
--- a/lib/csv/row.rb
+++ b/lib/csv/row.rb
@@ -19,7 +19,7 @@ class CSV
     #
     # The optional +header_row+ parameter can be set to +true+ to indicate, via
     # CSV::Row.header_row?() and CSV::Row.field_row?(), that this is a header
-    # row.  Otherwise, the row assumes to be a field row.
+    # row. Otherwise, the row assumes to be a field row.
     #
     # A CSV::Row object supports the following Array methods through delegation:
     #
--- a/lib/csv/row.rb
+++ b/lib/csv/row.rb
@@ -4,7 +4,7 @@ require "forwardable"
 
 class CSV
   #
-  # A CSV::Row is part Array and part Hash.  It retains an order for the fields
+  # A CSV::Row is part Array and part Hash. It retains an order for the fields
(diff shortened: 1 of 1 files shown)


lib/csv/row.rb
```diff
@@ -172,8 +172,8 @@ def []=(*args)
     #   <<( header_and_field_hash )
     #
     # If a two-element Array is provided, it is assumed to be a header and field
-    # and the pair is appended.  A Hash works the same way with the key being
-    # the header and the value being the field.  Anything else is assumed to be
+    # and the pair is appended. A Hash works the same way with the key being
+    # the header and the value being the field. Anything else is assumed to be
     # a lone field which is appended with a +nil+ header.
```
vbrazo: ```Anything else is assumed to be a lone field which is appended with a +nil+ header.```

Is `lone field` correct? I was wondering if it should be `line field`.

lib/csv/row.rb
```diff
@@ -13,13 +13,13 @@ class CSV
...
     #
-    # Construct a new CSV::Row from +headers+ and +fields+, which are expected
-    # to be Arrays.  If one Array is shorter than the other, it will be padded
+    # Constructs a new CSV::Row from +headers+ and +fields+, which are expected
+    # to be Arrays. If one Array is shorter than the other, it will be padded
     # with +nil+ objects.
     #
     # The optional +header_row+ parameter can be set to +true+ to indicate, via
     # CSV::Row.header_row?() and CSV::Row.field_row?(), that this is a header
-    # row.  Otherwise, the row is assumes to be a field row.
```
vbrazo: grammar issue

lib/csv/row.rb
kou: `lone field` is correct: "being alone" in https://www.dictionary.com/browse/lone

But we can use other wording.

vbrazo: gotcha! Thanks for the link @kou 👍

PR #102, opened 2019-08-18
Add extra docs for CSV classes

vbrazo:


vbrazo:
@kou that's understandable. I'd suggest adding the responsibilities above the classes/methods and remove any extra detail that could explain how to use the objects.

Right now `lib/csv/parser.rb` is a relatively big class, it doesn't have docs and needs to be improved.

kou:
OK. Let's do so.

How about adding "Don't use this class directly. This is an internal class." or something too?

PR #109, opened 2019-10-31
Feature/improve performance

grzegorz-jakubiak:
I used `fasterer` gem to identify some of the performance issues in this gem. It's not a lot but they make the code faster and clearer.


kou:
Please don't include multiple changes in one pull request. Please put one change in one pull request.
Could you also show how much faster by the change in each separated pull request?

PR #110, opened 2019-11-01
Do not loop forever when skip_lines regexp matches zero length with anchors

crazymykl:
This started failing in v3.0.2.

There is a minimal repro case in the test.

### DIFF
--- a/lib/csv/parser.rb
+++ b/lib/csv/parser.rb
@@ -769,7 +769,7 @@ class CSV
     def skip_needless_lines
       return unless @skip_lines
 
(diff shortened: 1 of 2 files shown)


crazymykl:
Another approach that would work is to set `skip_blanks` if the `skip_lines` pattern matches an empty string during initialization; that would avoid changes in the hot loop.

kou:
Thanks.
I chose `eos?` in `skip_needless_lines?` because the same problem is happen by another `skip_needless_lines?` call with the original approach.

PR #116, opened 2019-12-25
Keep buffers before calling CSV::Parser#emit_row inside of #parse_quotable_robust

azmin8744:
## Probrem

`CSV#line` sometimes returns nil. It seems to be caused by empty lines within quoted columns.
(I have not been reproduce the error yet with a minimum CSV which contains a column to be caused. Sorry.)

## How to fix

In the `Parser#parse_quotable_loose`, the `@scanner.keep_drop` and the `@scanner.keep_start` are called before calling the `Parser#emit_row`. but there are no calls them the same way in the `Parser#parse_quotable_robust`.

- `Parser#parse_quotable_loose`
 https://github.com/ruby/csv/blob/398b3564c5b0de78552dfad1b4f26013bf21bdba/lib/csv/parser.rb#L869-L872
- `Parser#parse_quotable_robust`
https://github.com/ruby/csv/blob/398b3564c5b0de78552dfad1b4f26013bf21bdba/lib/csv/parser.rb#L896-L909

So I guess that calling the `@scanner.keep_drop` and the `@scanner.keep_start` as well in the `Parser#parse_quotable_robust`, it will be able to ensure getting a line through `Parser#last_line`.

### DIFF
--- a/lib/csv/parser.rb
+++ b/lib/csv/parser.rb
@@ -893,6 +893,8 @@ class CSV
         if parse_column_end
           row << value
         elsif parse_row_end
+          @scanner.keep_drop
+          @scanner.keep_start
           if row.empty? and value.nil?
             emit_row([], &block) unless @skip_blanks
           else
@@ -905,6 +907,8 @@ class CSV
         elsif @scanner.eos?
           break if row.empty? and value.nil?
           row << value
+          @scanner.keep_drop
+          @scanner.keep_start
           emit_row(row, &block)
           break
         else


kou:
Could you show a Ruby script that reproduces this case?

azmin8744:
## Ruby script that this case occurred

A minimum script that reproduces this case is below;

```rb
CSV.open(file) do |csv|
  csv.each do |row|
    if csv.line.nil?
      puts row.to_s # => some rows are reported
    end
  end
end
```

 I have run this on macOS 10.15.2, with Ruby 2.6.5.

`file` is a CSV file with UTF-8 encoding and CRLF (even newline in quoted columns).
While investigating rows which I've got nil through CSV#line, I found that all of these lines contains empty lines within quoted columns. but I've not got what caused by essentially yet.

kou:
Could you also provide a CSV file that reproduces this case?

azmin8744:
I'm sorry for the delay in my response.  And I'm sorry too for that I cannot provide a CSV file you as it is because it is held by the customer.

Through my investigation, I have found out some features about some rows which returns `nil` with `CSV#line`. These are like below;

- Rows are located in the last of chunks(read by [`CSV::InputScanner#read_chunk`](https://github.com/ruby/csv/blob/398b3564c5b0de78552dfad1b4f26013bf21bdba/lib/csv/parser.rb#L206))
- The chunk begins with a part of unquoted columns
- The chunk ends with CR of row-separating CRLF

So I have tried to reproduce this case with a file that contains some lines with same features, but it won't be fine. 
FYI, I'll show a script I used for generating a CSV file that has similar features above.

```rb
RECORD_SIZE = 6144
ALLOWABLE_CHARACTERS = ("0".."Z").uniq.push("\r\n")
COLUMN_SIZE_LIMIT = 2048
ROW_TERMINATOR = "\",\r\n"

(1..10).each do |num|
  content = "#{sprintf("%04d", num)},"
  column = ""
  while content.bytesize < RECORD_SIZE do
    content << "\""
    while column.bytesize < COLUMN_SIZE_LIMIT or content.bytesize + column.bytesize < RECORD_SIZE - ROW_TERMINATOR.bytesize do
      column << ALLOWABLE_CHARACTERS.sample
    end
    content << column
    content << "0" if num % 4 == 0 # add another byte at the end of the chunk
    content << ROW_TERMINATOR
  end
  puts content
end
```

Overall, I understood this pull request is a bad idea. it will destroy the behavior of `CSV#line`. So I want to withdraw it.

azmin8744:
> The chunk ends with CR of row-separating CRLF

From this feature,  I made a patch below, and I can get a true text by `CSV#line` . But I  wonder it's a good idea.

```diff
diff --git a/lib/csv/parser.rb b/lib/csv/parser.rb
index 924ca46..e1c56df 100644
--- a/lib/csv/parser.rb
+++ b/lib/csv/parser.rb
@@ -232,8 +232,12 @@ class CSV
           @last_scanner = @inputs.empty?
           true
         else
+          @cr ||= "\r".encode(@encoding)
           chunk = input.gets(nil, @chunk_size)
           if chunk
+            if chunk.end_with?(@cr)
+              chunk << input.gets(nil, 1)
+            end
             raise InvalidEncoding unless chunk.valid_encoding?
             @scanner = StringScanner.new(chunk)
             if input.respond_to?(:eof?) and input.eof?
```

PR #124, opened 2020-05-14
Enhanced Rdoc

BurdetteLamar:
I wanted to change the argument from +data+ to +string_or_io+, but that's a code change, not a doc change.   Let me know if that would have been ok.

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -955,10 +955,14 @@ class CSV
       end
     end
 
-    # Returns the new \Array created by parsing the first line of +line+
+    # :call-seq:
+    #   CSV.parse_line(string_or_io)
+    #   CSV.parse_line(string_or_io, **options)
+    #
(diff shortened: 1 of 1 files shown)


BurdetteLamar:
@kou, ready for review.

lib/csv.rb
```diff
@@ -1008,22 +1052,48 @@ def table(path, **options)
     end
   end
 
+  # :call-seq:
+  #   CSV.new(string_or_io)
+  #   CSV.new(string_or_io, **options)
```
kou, line 1057: How about the followings?

```ruby
CSV.new(string)
CSV.new(io)
CSV.new(string, **options)
CSV.new(io, **options)
```

BurdetteLamar: Of course.  I'll see to it.

PR #129, opened 2020-06-01
Add `undef: :replace` for `CSV.open`

koic:
This PR adds `undef: :replace` option for `CSV.open`.

`File.open` has `undef: :replace` option, but `CSV.open` does not.
It would be convenient if `CSV.open` could have a shortcut by having `undef: :replace` option.

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -960,6 +960,11 @@ class CSV
     #   open( filename, mode = "rb", **options )
     #   open( filename, **options )
     #
+    # possible options elements:
+    #   hash form:
+    #     :undef => :replace # replace undefined conversion
+    #     :replace => string # replacement string ("?" or "\uFFFD" if not specified)
+    #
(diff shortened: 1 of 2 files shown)


koic:
#90 has been opened for this issue but lacks activity.
I came across an opportunity to want to use this option with `CSV.open` and I've opened it as a new PR.

lib/csv.rb
```diff
@@ -1021,6 +1026,7 @@ def open(filename, mode="r", **options)
       # wrap a File opened with the remaining +args+ with no newline
       # decorator
       file_opts = {universal_newline: false}.merge(options)
+      file_opts.merge!(undef: options.delete(:undef), replace: options.delete(:replace))
```
kou, line 1029:
```suggestion
      options.delete(:undef)
      options.delete(:replace)
```

koic: I updated it. Thanks for your review!

kou:
Thanks!

PR #130, opened 2020-06-03
Add `invalid: :replace` for `CSV.open`

koic:
This PR adds `invalid: :replace` for `CSV.open`. It is a PR similar to #129.

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -962,8 +962,10 @@ class CSV
     #
     # possible options elements:
     #   hash form:
-    #     :undef => :replace # replace undefined conversion
-    #     :replace => string # replacement string ("?" or "\uFFFD" if not specified)
+    #     :invalid => nil      # raise error on invalid byte sequence (default)
+    #     :invalid => :replace # replace invalid byte sequence
+    #     :undef => :replace   # replace undefined conversion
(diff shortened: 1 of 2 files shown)


lib/csv/writer.rb
```diff
@@ -156,7 +156,7 @@ def quote(field)
         else
           field = String(field)  # Stringify fields
           # represent empty fields as empty quoted fields
-          if (@quote_empty and field.empty?) or @quotable_pattern.match?(field)
+          if (@quote_empty and field.empty?) or @quotable_pattern.match?(field.scrub)
```
kou, line 159: I don't think that this is a good approach.
If we have `\uFFFD` in `@quotable_pattern`, it matches all invalid characters.

We need to process this case carefully something like:

 ```suggestion
          if (@quote_empty and field.empty?) or (filed.valid_encoding? and @quotable_pattern.match?(field)
```

Could you create a separated pull request for this case and then we can back to this pull request?

koic: Thanks for your suggestion that definitely a safer approach. I think that is definitely a safe approach and suitable. I've opened #131.

koic: I rebased this PR with the latest mater branch.

kou:
Thanks!

PR #131, opened 2020-06-04
Fix an error for `CSV.open`

koic:
Follow up to https://github.com/ruby/csv/pull/130/files#r434885191.

This PR fixes `ArgumentError` for `CSV.open` when processing invalid byte sequence in UTF-8.


kou:
Thanks.

It seems that our benchmark doesn't show significant slow down:

Before: https://github.com/ruby/csv/runs/729090943?check_suite_focus=true
After: https://github.com/ruby/csv/pull/131/checks?check_run_id=736633688

(I want to compare benchmark results automatically...)

PR #134, opened 2020-06-08
Add argument-checks for ::new

BurdetteLamar:
Doc says "should be Arrays," but does not check.  If not, the error will manifest downstream, where it  can be harder to debug.
Plus, I don't want to write in the doc that "an exception is raised if headers or fields does respond to :size or :zip."

### DIFF
--- a/lib/csv/row.rb
+++ b/lib/csv/row.rb
@@ -4,34 +4,79 @@ require "forwardable"
 
 class CSV
   #
-  # A CSV::Row is part Array and part Hash. It retains an order for the fields
-  # and allows duplicates just as an Array would, but also allows you to access
-  # fields by name just as you could if they were in a Hash.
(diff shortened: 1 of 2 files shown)


esparta:
This is one of the difficult trade-off to review, it usually not a problem to add checking when the method is called once or a few times on the cycle. In the case of `CSV::Row` any new branch can harm the process since it may be called a lot for a given CSV will 1000s of rows.

A new benchmark may be helpful here (CSV has some on the benchmark folder), demonstrating how this new checks are a not-issue.

kou:
We don't check them explicitly.
Because normal users don't create `CSV::Row` manually. Normally, `CSV::Row` is created by the csv library.
And it will cause performance overhead.

BurdetteLamar:
Very well.

PR #136, opened 2020-06-11
Add document for CSV.instance

BurdetteLamar:
Moved delegated methods to new section, so it can be cited from more places than CSV.open.
Added examples for CSV.instance.

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -719,7 +719,7 @@ class CSV
     #   instance(io = $stdout, **options) {|csv| ... }
     #
     # Creates or retrieves cached \CSV objects.
-    # For arguments and options, see CSV::new.
+    # For arguments and options, see CSV.new.
     #
     # ---
     #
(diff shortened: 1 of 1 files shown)


lib/csv.rb
```diff
@@ -674,18 +712,47 @@ def initialize(message, line_number)
...
+    #   instance(io = $stdout, **options)
+    #   instance(string, **options) {|csv| ... }
+    #   instance(io = $stdout, **options) {|csv| ... }
     #
-    # This method will return a CSV instance, just like CSV::new(), but the
-    # instance will be cached and returned for all future calls to this method for
-    # the same +data+ object (tested by Object#object_id()) with the same
-    # +options+.
+    # Creates or retrieves cached \CSV objects.
+    # For arguments and options, see CSV::new.
```
kou, line 722:
```suggestion
    # For arguments and options, see CSV.new.
```

PR #137, opened 2020-06-13
Rdoc

BurdetteLamar:
Enhancements for open, parse, minor tweaks.

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -832,7 +832,7 @@ class CSV
     # Returns an integer, or, if there were no rows, +nil+.
     #
     # * Argument +path+, if given, must be the path to a file.
-    # * Argument +io+, if given, must be an \IO object opened for reading.
+    # :include: ../doc/argument_io.rdoc
     # * Argument +mode+, if given, must be a \File mode
     #   See {Open Mode}[IO.html#method-c-new-label-Open+Mode].
     # * Arguments <tt>**options</tt> must be keyword options.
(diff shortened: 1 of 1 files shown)


BurdetteLamar:
@kou, did not mean to merge this.  What to do now?

lib/csv.rb
```diff
@@ -864,18 +863,18 @@ def filter(input=nil, output=nil, **options)
     #   ["baz", "2"]
     #
     # Read rows from an \IO object:
-    #   CSV.foreach(io) {|row| p row } # => 21
+    #   CSV.foreach(File.open(path)) {|row| p row } # => 21
```
kou, line 866: We should not use no `close` `File.open` in example. It's a bad mannar:

```ruby
File.open(path) do |input|
  CSV.foreach(io) {row p row } # => 21
end
```

lib/csv.rb
```diff
@@ -1022,10 +1021,10 @@ def generate_line(row, **options)
 
     #
     # :call-seq:
-    #   open( filename, mode = "rb", **options ) { |faster_csv| ... }
-    #   open( filename, **options ) { |faster_csv| ... }
-    #   open( filename, mode = "rb", **options )
-    #   open( filename, **options )
+    #   open(file_path, mode = "rb", **options ) -> new_csv
+    #   open(io, mode = "rb", **options ) -> new_csv
+    #   open(file_path, mode = "rb", **options ) { |csv| ... } -> new_csv
```
kou, line 1026: It returns the block's return value not the yielded `csv` instance.

BurdetteLamar:
Ok, thanks.  Will study tomorrow.

PR #139, opened 2020-06-14
doc: fix return value of open {} and use File.open {}

BurdetteLamar:


BurdetteLamar:
@kou, Something's terribly wrong here.  The file doc/argument_io.rdoc is not here.  I  thought it was already merged, along with references to it.  How to resolve?

lib/csv.rb
```diff
@@ -863,7 +863,10 @@ def filter(input=nil, output=nil, **options)
     #   ["baz", "2"]
     #
     # Read rows from an \IO object:
-    #   CSV.foreach(File.open(path)) {|row| p row } # => 21
+    # File.open(path) do |file|
+    #   CSV.foreach(file) {|row| p row } # => 21
+    # end
```
kou, line 868: Should we need to indent this code?

BurdetteLamar: Done, thanks.

PR #141, opened 2020-06-15
Add headers cases to CSV.parse

BurdetteLamar:


lib/csv.rb
```diff
@@ -1131,8 +1131,8 @@ def open(filename, mode="r", **options)
 
     #
     # :call-seq:
-    #   parse(string) -> array_of_arrays
-    #   parse(io) -> array_of_arrays
+    #   parse(string) -> array_of_arrays or csv_table
+    #   parse(io) -> array_of_arrays or csv_table
```
kou, line 1135: How about adding the following new entries instead of changing the current entries?

```ruby
parse(string, headers: ..., **options) -> csv_table
parse(io, header: ..., **options) -> csv_table
```

BurdetteLamar: Can you do it?

BurdetteLamar:
@kou, I think this is ready.

kou:
Thanks.

PR #142, opened 2020-06-17
Added headers to RDoc for CSV.foreach

BurdetteLamar:
Added headers option in RDoc for CSV.foreach.
Also in RDoc for CSV.parse, moved the headers from the call to the data -- more  realistic?

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -825,6 +825,8 @@ class CSV
     # :call-seq:
     #   foreach(path, mode='r', **options) {|row| ... ) -> integer or nil
     #   foreach(io, mode='r', **options {|row| ... ) -> integer or nil
+    #   foreach(path, mode='r', headers: ..., **options) {|row| ... ) -> integer or nil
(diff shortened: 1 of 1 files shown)


lib/csv.rb
```diff
@@ -882,6 +886,33 @@ def filter(input=nil, output=nil, **options)
     #   warning: Unsupported encoding foo ignored
     #   warning: Unsupported encoding bar ignored
     #
+    # ====== With Option +headers+
+    #
+    # With option +headers+, returns each row as a CSV::Row object.
```
esparta, line 891: May be worthy to mention that `headers` requires a truly value to yield `CSV::Row` objects, in the other hand, a falsie  value will behave as no `headers` parameter: `CSV.foreach(path, headers: nil)`

BurdetteLamar: Thanks, @esparta.  I'll review this.  I didn't want to say "set," b/c that's vague.  Maybe "truthy value" and link to the section "Option headers."

PR #145, opened 2020-06-24
Organize files in doc/

BurdetteLamar:
Organizes .rdoc files in directory doc/ into directories:
- doc/
  - arguments
  - options
    - common
    - generating
    - parsing

There will be more .rdoc files for doc/arguments, and perhaps more for additional diriectories.

The files in doc/options/common need some work, but that will be in a separate PR.  (One topic per PR.)


kou:
Thanks.

PR #154, opened 2020-07-02
RDoc: summary lists for options

BurdetteLamar:


lib/csv.rb
```diff
@@ -1782,9 +1826,19 @@ def lineno
...
+  #   string = "foo,0\nbar,1\nbaz,2\n"
+  #   path = 't.csv'
+  #   File.write(path, string)
+  #   CSV.open(path) do |csv|
+  #     csv.each do |row|
+  #       p [csv.lineno, csv.line]
+  #     end
+  #   end
+  # Output:
+  #   [1, "foo,0\r\n"]
```
esparta, line 1839: Not sure if this is happening on a different operating system, but most of the time it should not add `\r` to the  lines read :
```ruby
[1, "foo,0\n"]
[2, "bar,1\n"]
[3, "baz,2\n"]
```

lib/csv.rb
```diff
@@ -1782,9 +1826,19 @@ def lineno
     end
   end
 
-  #
-  # The last row read from this file.
-  #
+  # Returns the row most recently parsed:
```
esparta, line 1829:
```suggestion
  # Returns the row most recently read:
```
This is probably just semantics, but in this case line has not been parsed yet, taking the example below, the difference per iteration in the loop make it clear:

|csv.line (just read) | row (parsed)|
|---|---|
|"foo,0\n"|["foo", "0"]|
|"bar,1\n"|["bar", "1"]|
| "baz,2\n"|["baz", "2"]|

BurdetteLamar: Changed to  "line most recently read" in next PR.

lib/csv.rb
BurdetteLamar: This is what I got directly from +irb+, which I am loath to disturb.  No doubt OS-dependent.

kou: Did you run it on Windows, right?
Could you use `File.binwrite` instead of `File.write` to avoid adding `\r` automatically on Windows?

BurdetteLamar: Or just take out the \r?

BurdetteLamar: @kou, willing to remove on your say-so.

kou: Removing the `\r` from the output is also OK.

BurdetteLamar: Fixed in next push.

BurdetteLamar:
I have 6 more \r\n, having to do with row_sep.  Remove them, too?

lib/csv.rb
BurdetteLamar: I have 6 more \r\n, having to do with row_sep. Remove them, too?

BurdetteLamar: @kou, thoughts?

kou: Yes.

BurdetteLamar: Fixed in next push.

kou:
Thanks.

PR #157, opened 2020-07-15
RDoc for converters

BurdetteLamar:
This PR had to be this big because of cross-referencing (links) and to avoid redundancy.

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -452,12 +452,20 @@ using CSV::MatchP if CSV.const_defined?(:MatchP)
 # You can use a _write_ _converter_ to modify values before writing.
 # - See {Write Converters}[#class-CSV-label-Write+Converters].
 #
-# There are three forms for specifying converters to \CSV methods:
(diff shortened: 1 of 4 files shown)


lib/csv.rb
```diff
@@ -440,54 +440,188 @@
...
+#   - A \Proc field converter.
+#   - An \Array of field converter names.
 #
-# Converter +:integer+ converts each field that +Integer()+ accepts:
+# Display:
+#   CSV::Converters.each_pair do |name, value|
+#     if value.kind_of?(Proc)
+#       p [name, value.class]
+#     else
+#      p [name, value]
```
kou, line 609:
```suggestion
#       p [name, value]
```

kou:
Thanks.

BurdetteLamar:
@kou, does this eventually get merged into ruby/ruby?

kou:
Yes.
But we need to consider where we should put separated documents in ruby/ruby before we put our changes to ruby/ruby.

PR #158, opened 2020-07-18
Enhanced RDoc for several methods

BurdetteLamar:
Methods:
- each
- read
- header_row?
- shift
- inspect

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -2261,7 +2261,8 @@ class CSV
   include Enumerable
 
   # :call-seq:
-  #   csv.each -> self
+  #   csv.each -> enumerator
+  #   csv.each {|row| ...}
   #
   # Calls the block with each successive row.
   # The data source must be opened for reading.
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -2068,7 +2068,7 @@ class CSV
   ### End Delegation ###
 
   # :call-seq:
-  #   csv.<< row
+  #   csv << row -> self
   #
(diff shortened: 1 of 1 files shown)


lib/csv.rb
```diff
@@ -2260,25 +2260,56 @@ def header_convert(name = nil, &converter)
 
   include Enumerable
 
+  # :call-seq:
+  #   csv.each -> self
```
kou, line 2264: Could you change the return value to "undefined"?

```suggestion
  #   csv.each -> enumerator
  #   csv.each {|row| ...}
```

lib/csv.rb
```diff
@@ -2260,25 +2260,56 @@ def header_convert(name = nil, &converter)
...
+  #   csv.each do |row|
+  #     p row
+  #   end
   def each(&block)
     parser_enumerator.each(&block)
   end
 
   # :call-seq:
-  #     read
+  #   csv.read -> array or csv_table
```
kou, line 2306: How about splitting each pattern?

```suggestion
  #   csv.read -> array
  #   csv.read -> csv_table (CSV::Table is better?)
```

BurdetteLamar: > What I have here is what I see in the doc for such as Hash and Array.  People have been careful to use something (csv_table) that is English, not code.  I once wrote foo.to_h -> Hash, and had someone point out that it's somewhat ambiguous:  class Hash or instance of Hash?

kou: I see.

PR #159, opened 2020-07-20
Bump minimum ruby version to 2.5.0

GabrielNagy:
A dependency to stringio was added to csv, which requires Ruby version >= 2.5. Bump the gemspec version accordingly.


kou:
Good catch!

PR #162, opened 2020-08-04
Enhanced RDoc for Table#[]

BurdetteLamar:


lib/csv/table.rb
```diff
@@ -365,14 +365,91 @@ def headers
...
+    # Counts backward from the last row if +n+ is negative:
+    #   table[-1] # => #<CSV::Row "Name":"baz" "Value":"2">
+    #
+    # Returns +nil+ if +n+ is too large or too small:
+    #   table[4] # => nil
+    #   table[-4] => nil
+    #
+    # Raises an exception if the access mode is <tt>:row</tt>
+    # and +n+ is not an
+    # {Integer-convertible object}[doc/implicit_conversion_rdoc.html#label-Integer-Convertible+Objects].
```
kou, line 396: We can't refer it by `doc/...` because we may generate document only for csv.
How about using https://docs.ruby-lang.org/en/master/implicit_conversion_rdoc.html ?

BurdetteLamar: Oh, excellent!  Done!

kou:
Thanks!

PR #163, opened 2020-08-09
Enhanced RDoc for [] and []=

BurdetteLamar:
I had not covered the ground with previous RDoc for #[], so have revised that.  Both #[] and #[]= are complex, and need careful case-by-case treatment.

### DIFF
--- a/lib/csv/table.rb
+++ b/lib/csv/table.rb
@@ -487,7 +487,7 @@ class CSV
     #
     # If <tt>range.start</tt> is larger than <tt>table.size</tt>,
     # returns an \Array of +nil+ values:
-    #   table[4..4] table[4..4]
+    #   table[4..4] # => [nil, nil, nil]
     #
     # ---
     #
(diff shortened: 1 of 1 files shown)


lib/csv/table.rb
```diff
@@ -432,9 +462,41 @@ def headers
...
+    #
+    # If <tt>range.end</tt> is negative, calculates the ending index from the end:
+    #   table[0..-1] # => [["foo", "0"], ["bar", "1"], ["baz", "2"]]
+    #
+    # If <tt>range.start</tt> is negative, calculates the starting index from the end:
+    #   table[-2..2] # => [["foo", "0"], ["bar", "1"], ["baz", "2"]]
+    #
+    # If <tt>range.start</tt> is larger than <tt>table.size</tt>,
+    # returns an \Array of +nil+ values:
+    #   table[4..4] table[4..4]
```
kou, line 490: Could you update the result value?

BurdetteLamar: Thanks for catching this.  Fixed in new push.

BurdetteLamar:
@kou, is this ready to merge?

kou:
Thanks!

PR #166, opened 2020-08-17
Enhanced RDoc for CSV::Table

BurdetteLamar:
Methods:
- to_a
- to_csv
- inspect

Also:
- Tuned up some of the call-seq.
- Removed references to *-convertible, per new guide: https://github.com/ruby/ruby/blob/master/doc/method_documentation.rdoc#argument-description-if-necessary-

(I know, I've violated the one-topic principle.)


BurdetteLamar:
@kou, here's another PR.  This completes all the methods in CSV::Table except #dig.  I have to wait on a change from JeremyEvans for that.

kou:
Thanks!

PR #173, opened 2020-09-06
Enhanced RDoc for CSV::Row

BurdetteLamar:
Methods:
- to_h
- to_csv
- dig
- inspect

### DIFF
--- a/lib/csv/row.rb
+++ b/lib/csv/row.rb
@@ -635,10 +635,21 @@ class CSV
       @row == other
     end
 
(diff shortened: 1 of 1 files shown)


kou:
Thanks!

BurdetteLamar:
@kou:  Do users more often use CSV with headers?  Or without?

kou:
I don't know but both of them are typical use cases.
There are many CSVs with/without headers in the real world.

PR #174, opened 2020-09-11
Fix CSV.filter to preserve headers

BurdetteLamar:


lib/csv.rb
```diff
@@ -1058,9 +1058,11 @@ def filter(input=nil, output=nil, **options)
         end
       end
       # build input and output wrappers
-      input  = new(input  || ARGF,    **in_options)
+      input  = parse(input  || ARGF, **in_options)
```
kou, line 1061: We should not use `parse` here because `parse` parses all input at once. It may consume large resource.

lib/csv.rb
```diff
@@ -1058,9 +1058,11 @@ def filter(input=nil, output=nil, **options)
...
       output = new(output || $stdout, **out_options)
-
+      if input.instance_of?(CSV::Table)
+        output << input.headers
+      end
```
kou, line 1065: This adds headers twice with the following case:

```ruby
  def test_filter_with_headers
    input = <<-CSV
foo,0
bar,1
baz,2
    CSV
    output = ""
    CSV.filter(input, output, headers: ["Name", "Value"], output_write_headers: true) do |row|
      row
    end
    assert_equal(input, output)
  end
```

test/csv/interface/test_read_write.rb
```diff
@@ -24,6 +24,20 @@ def test_filter
...
 
+  def test_filter_with_headers
+    input = <<-CSV
+Name,Value
+foo,0
+bar,1
+baz,2
+    CSV
+    output = ""
+    CSV.filter(input, output, headers: true, write_headers: true) do |row|
```
kou, line 35: We should use `output_write_headers:` here. If we use `write_headers:`, `input` is also rewritten.

BurdetteLamar:
OK, thanks.  I don't have a proper fix for this, but I do think it's a bug, and I think the test I've included is valid.  Do you have a good fix for this?

kou:
OK. I take over this.

kou:
Done.

PR #176, opened 2020-09-18
Organize page Recipes better

BurdetteLamar:
Makes it easier to understand what's here.  Also, make level-1 headers more specific.

This is forward-looking.  There will be more recipes, for example, about Parsing.

### DIFF
--- a/doc/csv/recipes.rdoc
+++ b/doc/csv/recipes.rdoc
@@ -1,33 +1,46 @@
 == Recipes
 
 === Contents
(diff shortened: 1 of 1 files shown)


kou:
Thanks!
(It may be better that we split the file instead of using deep nested sections.)

BurdetteLamar:
I don't think it will be needed to nest deeper than this.  But I do think it will be important for all recipes to be listed (and linked) in one place, so that readers can find  the recipes they want.

PR #177, opened 2020-09-19
Recipes for field converters

BurdetteLamar:


doc/csv/recipes.rdoc
```diff
@@ -152,6 +156,51 @@ Output:
...
+  CSV::Converters[:strip] = strip_converter
+
+Without the new converter:
+  string = " foo , 0 \n bar , 1 \n baz , 2 \n"
+  array = CSV.parse(string)
+  array # => [[" foo ", " 0 "], [" bar ", " 1 "], [" baz ", " 2 "]]
+
+With the new converter:
+  string = " foo , 0 \n bar , 1 \n baz , 2 \n"
+  array = CSV.parse(string, converters: strip_converter)
```
kou, line 201: We don't need to register to `CSV::Converters` if we specify the custom converter proc to `converters:`:

```ruby
require "csv"

strip_converter = proc {|field| field.strip }
string = " foo , 0 \n bar , 1 \n baz , 2 \n"
array = CSV.parse(string, converters: strip_converter)
p array
# [["foo", "0"], ["bar", "1"], ["baz", "2"]]

Comment (kou) [reply doc/csv/recipes.rdoc:201]: If we register to `CSV::Converters`, we can specify the registered key instead of the proc:

```ruby
require "csv"

strip_converter = proc {|field| field.strip }
CSV::Converters[:strip] = strip_converter
string = " foo , 0 \n bar , 1 \n baz , 2 \n"
array = CSV.parse(string, converters: :strip)
p array
# [["foo", "0"], ["bar", "1"], ["baz", "2"]]
```

BurdetteLamar: Fixed (as unregistered) in next push.

Converter registration needs clarification over in lib/csv.rb.  That's for another PR.

BurdetteLamar: Added snippet to show registration.  Link to lib/csv.rb no longer needed.

kou:
Thanks!

PR #182, opened 2020-10-08
Refine RDoc for headers recipes

BurdetteLamar:
The idea is that each recipe should be very specific, and deal with only one task.


BurdetteLamar:
Aside:  @kou, is there anyplace online where the Rdoc (html) for master is available?

kou:
We don't have it yet.
We'll be able to prepare https://ruby.github.io/csv/ for it. But I can't work on it in a few months...

PR #183, opened 2020-10-09
More on Recipes headers

BurdetteLamar:
Make clear which headers (and corresponding TOC links) point to actual recipes:
- Header for actual recipe begins 'Recipe: ' and uses imperative voice verb.
- Header for non-recipe (covering section) uses gerund verb form.


kou:
Thanks!

PR #192, opened 2020-11-12
Removed needless editorconfig file

hsbt:
I found `.editorconfig` file under the `lib/csv` directory. It's not affected in this repository. So, It should be removed.


kou:
Good catch!

PR #198, opened 2020-12-29
Allow :newline option to CSV.open

nobu:
As `:newline` has priority over `:universal_newline` etc. in `IO.open`, so deal with both.
On the other hand, such options cannot be passed to `CSV#initialize`.

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1430,7 +1430,7 @@ class CSV
       options.delete(:invalid)
       options.delete(:undef)
       options.delete(:replace)
-      options.delete_if {|k, _| /newline\z/ =~ k}
+      options.delete_if {|k, _| /newline\z/.match?(k)}
 
       begin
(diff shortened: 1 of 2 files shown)


lib/csv.rb
```diff
@@ -1423,10 +1423,14 @@ def generate_line(row, **options)
...
       # decorator
-      file_opts = {universal_newline: false}.merge(options)
+      file_opts = options.dup
+      unless file_opts.key?(:newline)
+        file_opts[:universal_newline] ||= false
+      end
       options.delete(:invalid)
       options.delete(:undef)
       options.delete(:replace)
+      options.delete_if {|k, _| /newline\z/ =~ k}
```
esparta, line 1433: Since it only needs the true/false value, perhaps `.match?` is a better fit:
```suggestion
      options.delete_if {|k, _| /newline\z/.match?(k)}
```

kou:
Thanks!

PR #203, opened 2021-01-08
Enhaced RDoc for CSV::filter

BurdetteLamar:
Better RDoc for this versatile and powerful method.

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1010,8 +1010,14 @@ class CSV
     #   filter(in_string_or_io, out_string_or_io, **options) {|row| ... } -> array_of_arrays or csv_table
     #   filter(**options) {|row| ... } -> array_of_arrays or csv_table
     #
-    # Parses \CSV from a source (\String, \IO stream, or ARGF)
-    # and generates \CSV to an output (\String, \IO stream, or STDOUT).
+    # - Parses \CSV from a source (\String, \IO stream, or ARGF).
+    # - Calls the given block with each parsed row:
+    #   - Without headers, each row is an \Array.
+    #   - With headers, each row is a CSV::Row.
+    # - Generates \CSV to an output (\String, \IO stream, or STDOUT).
+    # - Returns the parsed source:
+    #   - Without headers, an \Array of \Arrays.
+    #   - With headers, a CSV::Table.
     #
     # When +in_string_or_io+ is given, but not +out_string_or_io+,
     # parses from the given +in_string_or_io+
@@ -1020,7 +1026,10 @@ class CSV
     # \String input without headers:
     #
     #   in_string = "foo,0\nbar,1\nbaz,2"
-    #   CSV::filter(in_string) {|row| row[0].upcase!; row[1] = - row[1].to_i }
+    #   CSV.filter(in_string) do |row|
+    #     row[0].upcase!
+    #     row[1] = - row[1].to_i
+    #   end # => [["FOO", 0], ["BAR", -1], ["BAZ", -2]]
     #
     # Output (to STDOUT):
     #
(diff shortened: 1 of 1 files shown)


BurdetteLamar:
Marking as draft.  Still a couple of things to do.

BurdetteLamar:
Now ready for review and merge.

lib/csv.rb
```diff
@@ -1006,60 +1006,149 @@ def instance(data = $stdout, **options)
...
-    # Example:
-    #   in_string = "foo,0\nbar,1\nbaz,2\n"
+    # When +in_string_or_io+ is given, but not +out_string_or_io+,
+    # parses from the given +in_string_or_io+
+    # and generates to STDOUT.
+    #
+    # \String input without headers:
+    #
+    #   in_string = "foo,0\nbar,1\nbaz,2"
+    #   CSV::filter(in_string) {|row| row[0].upcase!; row[1] = - row[1].to_i }
```
kou, line 1023: Why do you use multiple sentences in a line by `;`?
It may be difficult to read.

lib/csv.rb
```diff
@@ -1006,60 +1006,149 @@ def instance(data = $stdout, **options)
...
+    #
+    #   in_string = "Name,Value\nfoo,0\nbar,1\nbaz,2"
+    #   File.open('t.csv', 'w') do |out_io|
+    #     CSV.filter(in_string, out_io, headers: true) {|row| row[0].upcase!; row[1] = - row[1].to_i }
     #   end
-    #   out_string # => "FOO,0000\nBAR,1111\nBAZ,2222\n"
+    #   File.read('t.csv') # => "Name,Value\nFOO,0\nBAR,-1\nBAZ,-2\n"
+    #
+    # When neither +in_string_or_io+ nor +out_string_or_io+ given,
+    # parses from {ARGF}[https://ruby-doc.org/core/ARGF.html]
```
kou, line 1104: Could you use https://docs.ruby-lang.org/en/ instead of https://ruby-doc.org/ ?

lib/csv.rb
```diff
@@ -1006,60 +1006,149 @@ def instance(data = $stdout, **options)
...
+    # and generates to STDOUT.
+    #
+    # Without headers:
+    #
+    #   # Put Ruby code into a file.
+    #   ruby = <<-EOT
+    #     require 'csv'
+    #     CSV::filter {|row| row[0].upcase!; row[1] = - row[1].to_i }
```
kou, line 1112: Could you use `CSV.filter` style instead of `CSV::filter` style?

lib/csv.rb
```diff
@@ -1006,60 +1006,149 @@ def instance(data = $stdout, **options)
...
+    #   EOT
+    #   File.write('t.rb', ruby)
+    #   # Put some CSV into a file.
+    #   File.write('t.csv', "foo,0\nbar,1\nbaz,2")
+    #   # Run the Ruby code with CSV filename as argument.
+    #   `ruby t.rb t.csv`
```
kou, line 1118: Umm. This is a bit tricky to show `ARGF` behavior...

At least, we should use `system` instead of `` `...` ``:

```suggestion
    #   system(Gem.ruby, "t.rb", "t.csv")
```

lib/csv.rb
```diff
@@ -1006,60 +1006,149 @@ def instance(data = $stdout, **options)
...
+    #   FOO,0
+    #   BAR,-1
+    #   BAZ,-2
+    #
+    # With headers:
+    #
+    #   # Put Ruby code into a file.
+    #   ruby = <<-EOT
+    #     require 'csv'
+    #     p CSV::filter(headers: true) {|row| row[0].upcase!; row[1] = - row[1].to_i }
```
kou, line 1131:
```suggestion
    #     CSV::filter(headers: true) {|row| row[0].upcase!; row[1] = - row[1].to_i }
```

lib/csv.rb
BurdetteLamar: Fixed in ten places.

lib/csv.rb

lib/csv.rb

lib/csv.rb

lib/csv.rb
BurdetteLamar: Oops. Fixed.

BurdetteLamar:
Also:
- Added text at top to specify return value (Array of Arrays or CSV::Table) and passed row (Array or CSV::Row).
- Added return value in code examples, to emphasize difference between headers and not.

kou:
Thanks!

PR #207, opened 2021-02-23
Support pattern matching in CSV rows

kddeisz:


lib/csv/row.rb
```diff
@@ -659,8 +659,26 @@ def to_h
     end
     alias_method :to_hash, :to_h
 
+    # :call-seq:
+    #   row.deconstruct_keys(keys) -> hash
+    #
+    # Returns the new \Hash suitable for pattern matching containing only the
+    # keys specified as an argument.
+    def deconstruct_keys(keys)
+      keys.to_h { |key| [key, self[key]] }
```
baweaver, line 668: The keys, as supplied, will be `Symbol`s. This can present an issue as CSV rows will more than likely be `String` keyed:

```ruby
require 'csv'
# => true

csv = CSV.parse <<~RUBY, headers: true
  a,b,c
  1,2,3
  2,3,4
  4,5,6
RUBY
# => #<CSV::Table mode:col_or_row row_count:4>

csv.headers
# => ["a", "b", "c"]

csv.to_a
# => [["a", "b", "c"], ["1", "2", "3"], ["2", "3", "4"], ["4", "5", "6"]]

csv.first.class
# => CSV::Row

csv.first.fields
# => ["1", "2", "3"]
csv.first.headers
# => ["a", "b", "c"]
```

If you want to support this you would want to instead transform the keys to `String`, but this may be controversial as it conflates the two.

lib/csv/row.rb
```diff
@@ -659,8 +659,26 @@ def to_h
...
+
     alias_method :to_ary, :to_a
 
+    # :call-seq:
+    #   row.deconstruct -> array
+    #
+    # Returns the new \Array suitable for pattern matching containing the values
+    # of the row.
+    def deconstruct
+      @row.map(&:last)
```
baweaver, line 679: `CSV::Row#fields` may be better here.

lib/csv/row.rb
baweaver: This could be resolved one of a few ways.

In the first, retrieval of values would be done versus the `String` equivalent of the value:

```ruby
def deconstruct_keys(keys)
  keys.to_h { |k| [k, self[k.to_s]] }
end
```

This assumes, however commonly it may be, that all keys are `String` values. This could be resolved via:

```ruby
def deconstruct_keys(keys)
  first_header = headers.first
  case first_header
  when String then keys.to_h { |k| [k, self[k.to_s]] }
  when Symbol then keys.to_h { |k| [k, self[k]] }
  else {}
  end
end
```

...though this is far more verbose. I do not think that `Symbol` keys are commonly used in the context of `CSV`, so this concern might be ignored safely, but I leave that up to the core group of `CSV` to decide the risk.

kddeisz: I think I'd rather leave it as is for now. You can always pass `header_converters: :symbol` to the parse function, which would make this work as expected.

I think I would be really surprised if in my matching I specified symbols and it matched against strings.

run-test.rb
```diff
@@ -7,5 +7,8 @@
 $LOAD_PATH.unshift("lib")
 
 Dir.glob("test/csv/**/*test_*.rb") do |test_rb|
+  # Ensure we only load syntax that we can handle
+  next if RUBY_VERSION < "2.7" && test_rb.include?("test_patterns.rb")
```
kou, line 11:
```suggestion
  next if RUBY_VERSION < "2.7" && test_rb.end_with?("test_patterns.rb")
```

test/csv/test_patterns.rb
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# frozen_string_literal: true
+
+require_relative "helper"
+
+class TestCSVPatterns < Test::Unit::TestCase
```
kou, line 6:
```suggestion
class TestCSVPatternMatching < Test::Unit::TestCase
```

test/csv/test_patterns.rb
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
```
kou, line 1:
```suggestion
```

test/csv/test_patterns.rb
```diff
@@ -0,0 +1,23 @@
...
+# frozen_string_literal: true
+
+require_relative "helper"
+
+class TestCSVPatterns < Test::Unit::TestCase
+  extend DifferentOFS
+
+  def test_pattern_matching_hash
```
kou, line 9:
```suggestion
  def test_hash
```

test/csv/test_patterns.rb
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# frozen_string_literal: true
+
+require_relative "helper"
+
+class TestCSVPatterns < Test::Unit::TestCase
+  extend DifferentOFS
```
kou, line 7: We don't need this because we don't parse CSV in this test.

```suggestion
```

test/csv/test_patterns.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def test_pattern_matching_hash
+    case CSV::Row.new(%i{A B C}, [1, 2, 3])
+    in B: b, C: c
+      assert_equal(2, b)
+      assert_equal(3, c)
```
kou, line 13:
```suggestion
      assert_equal([2, 3], [b, c])
```

test/csv/test_patterns.rb
```diff
@@ -0,0 +1,23 @@
...
+# frozen_string_literal: true
+
+require_relative "helper"
+
+class TestCSVPatterns < Test::Unit::TestCase
+  extend DifferentOFS
+
+  def test_pattern_matching_hash
+    case CSV::Row.new(%i{A B C}, [1, 2, 3])
+    in B: b, C: c
```
kou, line 11: Could you add one more test case for `in B:, **rest`?

test/csv/test_patterns.rb
```diff
@@ -0,0 +1,23 @@
...
+      assert_equal(2, b)
+      assert_equal(3, c)
+    end
+  end
+
+  def test_pattern_matching_array
```
kou, line 17:
```suggestion
  def test_array
```

kou (commented):
Sorry for not responding this.
Could you add support for `in **rest` case?

test/csv/test_patterns.rb
kddeisz: Updated!

test/csv/test_patterns.rb
```diff
@@ -0,0 +1,28 @@
...
+    in B: b, C: c
+      assert_equal([2, 3], [b, c])
+    end
+  end
+
+  def test_hash_rest
+    case CSV::Row.new(%i{A B C}, [1, 2, 3])
+    in B: b, **rest
+      assert_equal(2, b)
+      assert_equal({ A: 1, C: 3 }, rest)
```
kou, line 18:
```suggestion
      assert_equal([2, { A: 1, C: 3 }], [b, rest])
```

kou:
Thanks!

PR #209, opened 2021-03-09
Implement unused #.empty? meth & clean up booleans

dainmiller:
Noticed this boolean was using a large amount of chained negative statements.

Figured a simpler implementation would be good for readers.

Sidenote: #.empty? method on this class was written but not implemented, but for this case is perf so used it.

image


dainmiller:
Fixing tests before re-opening

PR #210, opened 2021-03-09
Implement unused #.empty? meth & clean up booleans

dainmiller:
Noticed this boolean was using a large amount of chained negative statements.

Figured a simpler implementation would be good for readers.

Sidenote: #.empty? method on this class was written but not implemented, but for this case is perf so used it.


kou:
This is for your readability, right?

If it's right, I'll reject this request. This request evaluates `not (@nil_value.nil? and @empty_value_is_empty_string)` condition for each `#convert` call. It may have a bit negative performance impact. (We should confirm it by benchmark.)

PR #213, opened 2021-03-31
CI: Avoid Float 3.0 in YAML

olleolleolle:
Add quotes around "3.0" in YAML to avoid the issue reported in https://github.com/actions/runner/issues/849.


kou:
Thanks!

PR #216, opened 2021-06-01
Use test-unit gem instead of test-framework of ruby repo

hsbt:
I replaced test-framework provided by ruby/ruby repo to test-unit gem and the additional assertions named `CoreAssertions`.

This change makes unifying to assertion verb with `test-unit` gem.


kou:
Thanks.
I'll add some methods to test-unit gem to reduce more files.

hsbt:
@kou What can I do the next step? If you are okay to this, I will merge this.



> I'll add some methods to test-unit gem to reduce more files.

It's great for me.

kou:
Could you wait for a few days? I want to review all changes and try this on local. For example, I think that we should not remove `is_output_field_separator_deprecated` check in `test/lib/with_different_ofs.rb`. (I think that we need to check it with `$VERBOSE = true`.)

hsbt:
> Could you wait for a few days?

It's ok, no problem. 



> I think that we should not remove `is_output_field_separator_deprecated` check in `test/lib/with_different_ofs.rb`

👍  I picked this helper from ruby/ruby repo originally. It's better to adjust to CSV library.

kou:
* Fixed `WithDifferentOFS`. 
* Removed `CoreAssertions` entirely by adding needed assertions to test-unit.

hsbt:
Thanks!

PR #217, opened 2021-08-05
doc: Match text to the struct name

vreuter:


kou:
Good catch!

olleolleolle:
@vreuter Thanks for finding and fixing! 🎉

PR #218, opened 2021-08-20
Improve Ractor-compliance

rm155:
In order to become Ractor-compliant, CSV should avoid relying on class variables and global variables.

Additionally, since Converters and HeaderConverters are not accessible from Ractors, defining get_converters and get_header_converters (which access Ractor-local versions) can allow for an alternative within Ractors.


lib/csv.rb
```diff
@@ -879,7 +879,7 @@ def initialize(message, line_number)
   # This \Hash is intentionally left unfrozen, and may be extended with
   # custom field converters.
   # See {Custom Field Converters}[#class-CSV-label-Custom+Field+Converters].
-  Converters  = {
+  DefaultConverters  = {
```
kou, line 882: We can't accept this because this breaks backward compatibility.

lib/csv.rb
```diff
@@ -905,20 +905,49 @@ def initialize(message, line_number)
...
+  end
 
   # A \Hash containing the names and \Procs for the built-in header converters.
   # See {Built-In Header Converters}[#class-CSV-label-Built-In+Header+Converters].
   #
   # This \Hash is intentionally left unfrozen, and may be extended with
   # custom field converters.
   # See {Custom Header Converters}[#class-CSV-label-Custom+Header+Converters].
-  HeaderConverters = {
+  DefaultHeaderConverters = {
```
kou, line 929: We can't accept this because this breaks backward compatibility.

lib/csv.rb
```diff
@@ -995,8 +1024,13 @@ def instance(data = $stdout, **options)
...
       # fetch or create the instance for this signature
-      @@instances ||= Hash.new
-      instance = (@@instances[sig] ||= new(data, **options))
+      instances_hash = if defined?(Ractor)
+                         Ractor.current[:__CSV_instances_hash__] ||= Hash.new
+                       else
+                         @@instances ||= Hash.new
+                       end
+
+      instance = (instances_hash[sig] ||= new(data, **options))
```
kou, line 1033: Ractor users should not use `CSV.instance`.
They should create your own CSV instance.

lib/csv.rb
```diff
@@ -1187,7 +1221,8 @@ def instance(data = $stdout, **options)
     #   See {Options for Parsing}[#class-CSV-label-Options+for+Parsing].
     def filter(input=nil, output=nil, **options)
       # parse options for input, output, or both
-      in_options, out_options = Hash.new, {row_sep: $INPUT_RECORD_SEPARATOR}
+      row_sep = (defined?(Ractor) && (Ractor.current != Ractor.main)) ? "\n" : $INPUT_RECORD_SEPARATOR
+      in_options, out_options = Hash.new, {row_sep: row_sep}
```
kou, line 1225: Why do we need to avoid using `$INPUT_RECORD_SEPARATOR`?
It's sharable in Ractors because it's a frozen string.

lib/csv/parser.rb
```diff
@@ -476,12 +476,12 @@ def prepare_strip
...
-    begin
+    def string_scanner_scan_accept_string
       StringScanner.new("x").scan("x")
     rescue TypeError
-      @@string_scanner_scan_accept_string = false
+      false
     else
-      @@string_scanner_scan_accept_string = true
+      true
     end
```
kou, line 485: Can we use constant instead of class variable to avoid multiple `StringScanner.new.scan` check?

kou:
How did you confirm that these changes are needed to use in Ractor?
Can we create test cases for Ractor?

lib/csv.rb
```diff
@@ -905,20 +905,49 @@ def initialize(message, line_number)
     },
     all:       [:date_time, :numeric],
   }
+  Ractor.make_shareable(DefaultConverters) if defined?(Ractor)
+  Converters = DefaultConverters.dup
```
kou, line 909: Ah, this keeps backward compatibility.
But Ractor users can't use converters added by users.

rm155:
> 
> 
> How did you confirm that these changes are needed to use in Ractor?
> Can we create test cases for Ractor?

If CSV tries to parse a file from a Ractor, it looks for the converters, but since they are unfrozen shared constants, it can't access them.

I've written some tests for using CSV in Ractors. I can upload them soon.

lib/csv.rb
rm155: I believe this change basically makes CSV.instance find a Ractor-local instance. Could you explain more about why this might be a problem? Thanks!

lib/csv.rb
rm155: I agree that it is a frozen string, but according to the specifications $INPUT_RECORD_SEPARATOR can't be used outside the main Ractor. Unless it's outdated, I think that this was mentioned here (as $/):   https://bugs.ruby-lang.org/issues/17268. I hope I've understood correctly!

lib/csv/parser.rb
rm155: Yes, I see what you mean. I can make that change soon. Thanks!

lib/csv.rb
rm155: I agree that if a user defines a converter in the main Ractor, then the converter can't automatically be used from non-main-Ractors. This change would allow users to define converters within one Ractor and use them just within the same Ractor. This can be done by getting the Ractor-local converters hash with get_converters, and then adding to it.

Without the change, converters are completely inaccessible from Ractors, so I think having Ractor-local versions is a step forward. Since it does not currently seem possible to have shared hashes among Ractors, I am not sure how the ideal situation could be achieved.

lib/csv.rb
kou: Yes. `CSV.instance` can work in Ractor with this change.
It's just API design decision that we don't recommend (support) to use `CSV.instance` in Ractor.
We can just add "This API isn't Ractor-safe" to document here.

rm155: Okay, I understand!

test/csv/helper.rb
```diff
@@ -4,6 +4,8 @@
 require "csv"
 
 require_relative "../lib/with_different_ofs"
+require_relative "../lib/core_assertions"
```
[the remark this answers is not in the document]
rm155: Sure, thanks!

lib/csv.rb
kou: Ah, https://bugs.ruby-lang.org/issues/18129 ?

kou: `$/` is deprecated since Ruby 3.0: https://bugs.ruby-lang.org/issues/14240
We may be able to replace `$INPUT_RECORD_SEPARATOR` by '"\n"' with Ruby 3.0 or later.

test/csv/helper.rb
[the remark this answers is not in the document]
kou: Could you use this change for tests that use Ractor?

```diff
diff --git a/csv.gemspec b/csv.gemspec
index 6948e9a..15ffc49 100644
--- a/csv.gemspec
+++ b/csv.gemspec
@@ -60,5 +60,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "bundler"
   spec.add_development_dependency "rake"
   spec.add_development_dependency "benchmark_driver"
-  spec.add_development_dependency "test-unit", ">= 3.4.3"
+  spec.add_development_dependency "test-unit", ">= 3.4.6"
 end
diff --git a/test/csv/interface/test_read.rb b/test/csv/interface/test_read.rb
index 62ddae2..4ec40ec 100644
--- a/test/csv/interface/test_read.rb
+++ b/test/csv/interface/test_read.rb
@@ -32,6 +32,24 @@ class TestCSVInterfaceRead < Test::Unit::TestCase
     assert_equal(@rows, rows)
   end
 
+  if respond_to?(:ractor)
+    ractor
+    def test_foreach_in_ractor
+      ractor = Ractor.new(@input.path) do |path|
+        rows = []
+        CSV.foreach(path, col_sep: "\\t", row_sep: "\\r\\n").each do |row|
+          rows << row
+        end
+        Ractor.yield rows
+      end
+      rows = [
+        ["1", "2", "3"],
+        ["4", "5"],
+      ]
+      assert_equal(rows, ractor.take)
+    end
+  end
+
   def test_foreach_mode
     rows = []
     CSV.foreach(@input.path, "r", col_sep: "\t", row_sep: "\r\n").each do |row|
```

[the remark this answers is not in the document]
rm155: Okay, I have tried using this change instead of assert_ractor. It seems to work, but please let me know if I should change it in any way. Thanks!

lib/csv.rb
kou: Done on master.

lib/csv.rb
kou: I decided API for `Converters` and `HeaderConverters`:

* If a Ractor user doesn't use any built-in converters, the user can use `CSV`. (committed)
* If a Ractor user wants to use any built-in converters in non-main-Ractors, the user must call `Ractor.make_sharable(CSV::Converters)` and `Ractor.make_sharedable(CSV::HeaderConverters)` after the user adds additional converters and before non-main-Ractors are created.
  * This should be documented. (Could you work on it in this pull request?)

lib/csv.rb
kou: Could you document it in this pull request?

lib/csv.rb
kou: Could you rebase on master?
You need to resolve conflict. Sorry.

test/csv/interface/test_read.rb
```diff
@@ -32,6 +32,24 @@ def test_foreach
     assert_equal(@rows, rows)
   end
 
+  if(respond_to? :ractor)
```
kou, line 35:
```suggestion
  if respond_to?(:ractor)
```

test/csv/interface/test_read.rb
```diff
@@ -250,6 +268,20 @@ def test_read
                  CSV.read(@input.path, col_sep: "\t", row_sep: "\r\n"))
   end
 
+  if(respond_to? :ractor)
```
kou, line 271:
```suggestion
  if respond_to?(:ractor)
```

test/csv/interface/test_write.rb
```diff
@@ -25,6 +25,22 @@ def test_generate_default
     CSV
   end
 
+  if(respond_to? :ractor)
```
kou, line 28:
```suggestion
  if respond_to?(:ractor)
```

test/csv/interface/test_write.rb
```diff
@@ -101,6 +117,25 @@ def test_append_row
     CSV
   end
 
+
+  if(respond_to? :ractor)
```
kou, line 121:
```suggestion
  if respond_to?(:ractor)
```

rm155:
Thank you so much for your suggestions! I will update the pull request and rebase within the next few days.

rm155:
I have made the updates and done the rebasing. I hope it is okay! Please let me know if there is any way I can improve it.

kou:
I'll merge this.
Thanks!

rm155:
> 
> 
> I'll merge this. Thanks!

Thanks so much for your support and patience!

PR #220, opened 2021-10-01
Fix for parser getting confused after faulty row with CRLF line ending. 

anakinj:
It looks like the parser is currently not able to fully ignore a broken line because of the regex used for detecting the line ending in the file. The current behaviour will move the parser to the first `\r` char in the line. But for files using CRLF as line endings that is not enough.

This PR changes the detection to consider the combinations `\r\n` or `\r` or `\n` when ignoring a broken line.

### DIFF
--- a/lib/csv/parser.rb
+++ b/lib/csv/parser.rb
@@ -527,7 +527,7 @@ class CSV
 
       @cr = "\r".encode(@encoding)
       @lf = "\n".encode(@encoding)
-      @cr_or_lf = Regexp.new("[\r\n]".encode(@encoding))
+      @line_end = Regexp.new("\r\n|\n|\r".encode(@encoding))
       @not_line_end = Regexp.new("[^\r\n]+".encode(@encoding))
     end
 
@@ -915,7 +915,7 @@ class CSV
             message = "Any value after quoted field isn't allowed"
             raise MalformedCSVError.new(message, @lineno)
           elsif @unquoted_column_value and
-                (new_line = @scanner.scan(@cr_or_lf))
+                (new_line = @scanner.scan(@line_end))
             ignore_broken_line
             message = "Unquoted fields do not allow new line " +
                       "<#{new_line.inspect}>"
@@ -924,7 +924,7 @@ class CSV
             ignore_broken_line
             message = "Illegal quoting"
             raise MalformedCSVError.new(message, @lineno)
-          elsif (new_line = @scanner.scan(@cr_or_lf))
+          elsif (new_line = @scanner.scan(@line_end))
             ignore_broken_line
             message = "New line must be <#{@row_separator.inspect}> " +
                       "not <#{new_line.inspect}>"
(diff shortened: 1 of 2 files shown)


test/csv/parse/test_row_separator.rb
```diff
@@ -13,4 +13,17 @@ def test_multiple_characters
...
+    data = <<-CSV
+"1","OK"\r
+"2",""NOT" OK"\r
+"3","OK"\r
+CSV
+    csv = CSV.new(data)
+
+    assert_equal('1',  csv.shift[0])
+    assert_raise(CSV::MalformedCSVError) { csv.shift }
+    assert_equal('3',  csv.shift[0])
```
olleolleolle, line 27: It's as if there's an additional space after the first string in the '1' and '3' assertions.

olleolleolle (commented):
Cool improvement.

test/csv/parse/test_row_separator.rb
anakinj: Good catch. Should be fixed now.

lib/csv/parser.rb
```diff
@@ -527,7 +527,7 @@ def prepare_separators
 
       @cr = "\r".encode(@encoding)
       @lf = "\n".encode(@encoding)
-      @cr_or_lf = Regexp.new("[\r\n]".encode(@encoding))
+      @cr_or_lf = Regexp.new("\r\n|\n|\r".encode(@encoding))
```
kou, line 530: `@cr_or_lf` isn't match with this change. Could you rename this variable?

```suggestion
      @line_end = Regexp.new("\r\n|\n|\r".encode(@encoding))
```

test/csv/parse/test_row_separator.rb
```diff
@@ -13,4 +13,17 @@ def test_multiple_characters
                    CSV.parse("a\r\nb\r\n", row_sep: "\r\n"))
     end
   end
+
+  def test_skip_row_recovery_after_faulty_quoting_with_row_ending_with_cr_lf
```
kou, line 17: This file is for `row_sep` option. Could you move this to `test/csv/parse/test_invalid.rb`?

We can simplify test name like `test_ignore_invalid_line_cr_lf`.

test/csv/parse/test_row_separator.rb
```diff
@@ -13,4 +13,17 @@ def test_multiple_characters
...
+    data = <<-CSV
+"1","OK"\r
+"2",""NOT" OK"\r
+"3","OK"\r
+CSV
+    csv = CSV.new(data)
+
+    assert_equal('1', csv.shift[0])
```
kou, line 25:
```suggestion
    assert_equal(["1", "OK"], csv.shift)
```

test/csv/parse/test_row_separator.rb
```diff
@@ -13,4 +13,17 @@ def test_multiple_characters
...
+    assert_raise(CSV::MalformedCSVError) { csv.shift }
+    assert_equal('3', csv.shift[0])
```
kou, line 27:
```suggestion
    assert_equal(["3", "OK"], csv.shift)
```

anakinj:
Thanks for the feedback @kou. PR updated accordingly.

kou:
Thanks.
I've merged.

PR #221, opened 2021-10-07
Fix case where an \r\n row separator will be split when reading a chunk

jeremyevans:
In this case, read one more character.

This is a suboptimal fix, as it doesn't fix handling of row
separators that aren't two characters and starting with \r.
A better fix would handle all multibyte row separators.
However, as \r\n is one of the most common row separators,
I think it's useful to merge this until a more generic
solution is developed.

Fixes [Bug #18245]

### DIFF
--- a/lib/csv/parser.rb
+++ b/lib/csv/parser.rb
@@ -235,6 +235,9 @@ class CSV
         else
           chunk = input.gets(nil, @chunk_size)
           if chunk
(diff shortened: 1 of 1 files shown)


kou:
I pushed the row separator fix. It fixes a problem with the reported data. But it's not a real fix.
We should fix it later: #230

PR #224, opened 2021-11-03
Fix typos [ci skip]

nobu:


olleolleolle (approved):
Looks correct!

kou:
Good catch!

Issue #225, opened 2021-11-17
A placeholder error message is printed when parsing files with ambiguous options


adamroyjones:
I've noticed that the csv gem produces a placeholder error message when attempting to parse a TSV with ambiguous parsing options.

If one were to parse the string

```
foo\t\tbar
```

when directed to strip whitespace (i.e., to treat whitespace as insignificant) and use tabs as field separators (i.e., to treat whitespace as significant), then should one parse it as `["foo\t\tbar"]`, `["foo", "bar"]`, or `["foo", "", "bar"]`? There's no way to choose a parsing strategy that wouldn't cause a reasonable surprise to someone.

I think an error should be produced that informs the user that this is a problem; at the moment, this doesn't happen. With Ruby 3.0.2 and csv 3.2.1, the file

```ruby
require "csv"

sep = "\t"
tsv_file = "example.tsv"

File.open(tsv_file, "w") { |f| f.puts("foo#{sep}bar") }
CSV.read(tsv_file, col_sep: sep, strip: true)
```

produces the error

```
csv-3.2.1/lib/csv/parser.rb:935:in `parse_quotable_robust': TODO: Meaningful message in line 1. (CSV::MalformedCSVError)
```

I think this is misleading; the file is not really "malformed".

Do you agree with the above? If so, I'm happy to offer up a merge request around this (e.g. to check the options before beginning the parsing, to make sure that if `strip` is set to `true` then `col_sep` must not be whitespace).

kou:
Good catch!
Could you open a pull request?

adamroyjones:
Thanks for getting back to me so quickly—I'm glad I'm not going mad and that you agree!

I'll work on it and raise a merge request soon.

PR #226, opened 2021-11-17
Add handling for ambiguous parsing options

adamroyjones:
This pull request addresses [this bug report](https://github.com/ruby/csv/issues/225).

Let me know if there are any issues or places where I can improve it!

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -366,8 +366,9 @@ using CSV::MatchP if CSV.const_defined?(:MatchP)
 # - +header_converters+: Specifies the header converters to be used.
 # - +skip_blanks+: Specifies whether blanks lines are to be ignored.
 # - +skip_lines+: Specifies how comments lines are to be recognized.
-# - +strip+: Specifies whether leading and trailing whitespace are
-#   to be stripped from fields.
+# - +strip+: Specifies whether leading and trailing whitespace are to be
+#   stripped from fields. This must be compatible with +col_sep+; if it is not,
+#   then an +ArgumentError+ exception will be raised.
 # - +liberal_parsing+: Specifies whether \CSV should attempt to parse
 #   non-compliant data.
 # - +nil_value+: Specifies the object that is to be substituted for each null (no-text) field.
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -341,6 +341,7 @@ using CSV::MatchP if CSV.const_defined?(:MatchP)
 #     liberal_parsing:    false,
 #     nil_value:          nil,
 #     empty_value:        "",
+#     strip:              false,
 #     # For generating.
 #     write_headers:      nil,
 #     quote_empty:        true,
@@ -348,7 +349,6 @@ using CSV::MatchP if CSV.const_defined?(:MatchP)
 #     write_converters:   nil,
 #     write_nil_value:    nil,
 #     write_empty_value:  "",
-#     strip:              false,
 #   }
 #
 # ==== Options for Parsing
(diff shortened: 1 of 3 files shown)


lib/csv.rb
```diff
@@ -854,6 +855,10 @@ def initialize(message, line_number)
     end
   end
 
+  # The error thrown when the parser is given mutually incompatible options.
+  class InvalidOptionsError < RuntimeError
+  end
```
kou, line 860: Could you use the builtin `ArgumentError` instead of creating a new error class? It's suitable for this case.

lib/csv/parser.rb
```diff
@@ -531,6 +532,31 @@ def prepare_separators
...
+      end
+
+      space = [" ", "\t", "\f", "\v"]
+      default_condition = !@strip.is_a?(String) &&
+        (@column_separator.start_with?(*space) || @column_separator.end_with?(*space))
+      if default_condition
+        raise InvalidOptionsError,
+              "The provided strip (true) and " \
+              "col_sep (#{@escaped_column_separator}) options are incompatible."
+      end
```
kou, line 557: Can we simplify this like the following?

```ruby
if @strip.is_a?(String)
  if @column_separator.start_with?(@strip) || @column_separator.end_with?(@strip)
    raise ...
  end
else
  if Regexp.new("\\A[#{@escaped_strip}]|[#{@escaped_strip}]\\z").match?(@column_separator)
    raise ...
  end
end

File: test/csv/parse/test_strip.rb [hunk 3]
```diff
@@ -80,4 +80,33 @@ def test_do_not_strip_crlf
                            %Q{"a" ,"b " \r\n},
                            strip: true))
   end
+
+  def test_raise_error_if_col_sep_is_tab_and_strip_is_true
```
kou, line 84: Can we simplify this like the following?

```suggestion
  def test_col_sep_incompatible_true
```

This file is for `:strip`. So we don't need to mention `strip` in this file.

test/csv/parse/test_strip.rb
```diff
@@ -80,4 +80,33 @@ def test_do_not_strip_crlf
...
+    message = "The provided strip (true) and " \
+              "col_sep (\\t) options are incompatible."
+    assert_raise_with_message(CSV::InvalidOptionsError, message) do
+      CSV.parse_line(%Q{"a"\t"b"\n},
+                     col_sep: "\t",
+                     strip: true)
+    end
+  end
+
+  def test_raise_error_if_col_sep_is_tab_and_strip_is_tab
```
kou, line 94:
```suggestion
  def test_col_sep_incompatible_string
```

test/csv/parse/test_strip.rb
```diff
@@ -80,4 +80,33 @@ def test_do_not_strip_crlf
...
+    message = "The provided strip (\\t) and " \
+              "col_sep (\\t) options are incompatible."
+    assert_raise_with_message(CSV::InvalidOptionsError, message) do
+      CSV.parse_line(%Q{"a"\t"b"\n},
+                     col_sep: "\t",
+                     strip: "\t")
+    end
+  end
+
+  def test_do_not_raise_error_if_col_sep_and_strip_are_compatible
```
kou, line 104:
```suggestion
  def test_col_sep_compatible_string
```

lib/csv.rb
adamroyjones: Yes, you're quite right—I've now updated the pull request.

lib/csv/parser.rb
adamroyjones: Yes, I think so. Updated.

kou:
Thanks!

adamroyjones:
> Thanks!

Any time! This was all really pleasant to work on—the Ruby community's reputation is deserved. Thanks!

PR #229, opened 2021-12-07
Add contents summary to CSV::Table#inspect

ericgpks:
I thought that it is useful for us if this library has a function to return some rows results as a table.
So changed the method to return a table with a header and rows.

### DIFF
--- a/lib/csv/table.rb
+++ b/lib/csv/table.rb
@@ -1000,9 +1000,7 @@ class CSV
     # (see {Option +write_headers+}[../CSV.html#class-CSV-label-Option+write_headers]):
     #   table.to_csv(write_headers: false) # => "foo,0\nbar,1\nbaz,2\n"
     #
-    # Limit rows if option +limit+ is given like +2+
-    #   source = "Name,Value\nfoo,0\nbar,1\nbaz,2\n"
-    #   table = CSV.parse(source, headers: true)
+    # Limit rows if option +limit+ is given like +2+:
     #   table.to_csv(limit: 2) # => "Name,Value\nfoo,0\nbar,1\n"
     def to_csv(write_headers: true, limit: nil, **options)
       array = write_headers ? [headers.to_csv(**options)] : []
--- a/lib/csv/table.rb
+++ b/lib/csv/table.rb
@@ -1028,14 +1028,17 @@ class CSV
     end
 
     # :call-seq:
-    #   table.inspect(row_counts) -> self
+    #   table.inspect() => string
     #
-    # Return table with specified rows data with header
+    # Returns a table with specified rows data with header
     #   source = "Name,Value\nfoo,0\nbar,1\nbaz,2\n"
     #   table = CSV.parse(source, headers: true)
-    #   table.inspect(2) # => #<CSV::Row "Name":"foo" "Value":"0">, #<CSV::Row "Name":"bar" "Value":"1">
     #
-    def inspect(row_counts)
+    # Example:
+    #   table.inspect() # => #<CSV::Row "Name":"foo" "Value":"0">, #<CSV::Row "Name":"bar" "Value":"1">
+    #
+    def inspect()
+      row_counts = 5
       array = [headers.to_csv]
       if row_counts.nil?
         return array.join("")
--- a/lib/csv/table.rb
+++ b/lib/csv/table.rb
@@ -999,9 +999,11 @@ class CSV
     # Omits the headers if option +write_headers+ is given as +false+
     # (see {Option +write_headers+}[../CSV.html#class-CSV-label-Option+write_headers]):
     #   table.to_csv(write_headers: false) # => "foo,0\nbar,1\nbaz,2\n"
-    def to_csv(write_headers: true, **options)
+    def to_csv(write_headers: true, limit: nil, **options)
       array = write_headers ? [headers.to_csv(**options)] : []
-      @table.each do |row|
+      limit ||= @table.size
+      limit = @table.size + 1 + limit if limit < 0
+      @table.first(limit).each do |row|
         array.push(row.fields.to_csv(**options)) unless row.header_row?
       end
 
@@ -1028,7 +1030,7 @@ class CSV
     end
 
     # :call-seq:
-    #   table.inspect() => string
+    #   table.inspect => string
     #
     # Returns a table with specified rows data with header
     #   source = "Name,Value\nfoo,0\nbar,1\nbaz,2\n"
@@ -1037,16 +1039,11 @@ class CSV
     # Example:
     #   table.inspect() # => #<CSV::Row "Name":"foo" "Value":"0">, #<CSV::Row "Name":"bar" "Value":"1">
     #
-    def inspect()
-      row_counts = 5
-      array = [headers.to_csv]
-      if row_counts.nil?
-        return array.join("")
-      end
-      @table.first(row_counts).each do |row|
-        array.push(row.fields.to_csv) unless row.header_row?
-      end
-      array.join("")
+    def inspect
+      inspected = +"#<#{self.class} mode:#{@mode} row_count:#{to_a.size}>"
+      summary = to_csv(limit: 5)
+      inspected << "\n" << summary if summary.encoding.ascii_compatible?
+      inspected
     end
   end
 end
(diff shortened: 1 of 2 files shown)


esparta:
I don't think is a good idea to change the signature and the return value of `Table#inspect` just to have a peek of data. If the user need that it may be a better idea to use `Table#values_at(<range>)`, tho.

With this change there's a high risk some old scripts suddenly start to fail, some others (probably less) that was parsing or grep-ing the result will also fail.

lib/csv/table.rb
```diff
@@ -1028,19 +1028,22 @@ def dig(index_or_header, *index_or_headers)
...
-    # Example:
+    # Return table with specified rows data with header
     #   source = "Name,Value\nfoo,0\nbar,1\nbaz,2\n"
     #   table = CSV.parse(source, headers: true)
-    #   table.inspect # => "#<CSV::Table mode:col_or_row row_count:4>"
-    def inspect
-      "#<#{self.class} mode:#{@mode} row_count:#{to_a.size}>".encode("US-ASCII")
+    #   table.inspect(2) # => #<CSV::Row "Name":"foo" "Value":"0">, #<CSV::Row "Name":"bar" "Value":"1">
+    #
+    def inspect(row_counts)
```
kou, line 1038: Generally, `#inspect` should not require any arguments. It breaks compatibility and some tools such as irb can't work with this object.
How about using the constant `row_counts` such as 5 instead of receiving it from argument?

lib/csv/table.rb
```diff
@@ -1028,19 +1028,22 @@ def dig(index_or_header, *index_or_headers)
     end
 
     # :call-seq:
-    #   table.inspect => string
+    #   table.inspect(row_counts) -> self
```
kou, line 1031: `#inspect` must returns `string`.

lib/csv/table.rb
```diff
@@ -1028,19 +1028,22 @@ def dig(index_or_header, *index_or_headers)
...
-    #
-    # Example:
+    # Return table with specified rows data with header
     #   source = "Name,Value\nfoo,0\nbar,1\nbaz,2\n"
     #   table = CSV.parse(source, headers: true)
-    #   table.inspect # => "#<CSV::Table mode:col_or_row row_count:4>"
-    def inspect
-      "#<#{self.class} mode:#{@mode} row_count:#{to_a.size}>".encode("US-ASCII")
+    #   table.inspect(2) # => #<CSV::Row "Name":"foo" "Value":"0">, #<CSV::Row "Name":"bar" "Value":"1">
+    #
```
kou, line 1037: Could you keep the current documentation outline?

```rdoc
:call-seq:
  ...

Returns ...

Example:
  ...
```

lib/csv/table.rb
```diff
@@ -1028,19 +1028,22 @@ def dig(index_or_header, *index_or_headers)
...
+    def inspect(row_counts)
+      array = [headers.to_csv]
+      if row_counts.nil?
+        return array.join("")
+      end
+      @table.first(row_counts).each do |row|
+        array.push(row.fields.to_csv) unless row.header_row?
+      end
+      array.join("")
```
kou, line 1046: How about "adding" some records to the current result instead of "replacing" the current result?

```text
#<CSV::Table mode:col_or_row row_count:4>
Name,Value
foo,0
bar,1
```

instead of

```text
Name,Value
foo,0
bar,1
```

test/csv/test_table.rb
```diff
@@ -552,20 +552,22 @@ def test_array_delegation
...
-    assert_include(str, "mode:#{@table.mode}", "Mode not shown.")
+  def test_inspect_nil
+    row_counts = nil
+    table = @table.inspect(row_counts)
+    assert_equal(<<-CSV, table)
+A,B,C
+    CSV
   end
 
-  def test_inspect_encoding_is_ascii_compatible
```
kou, line 564: Could you keep the current tests to keep backward compatibility?

lib/csv/table.rb
```diff
@@ -1028,19 +1028,22 @@ def dig(index_or_header, *index_or_headers)
...
+    #   table.inspect(2) # => #<CSV::Row "Name":"foo" "Value":"0">, #<CSV::Row "Name":"bar" "Value":"1">
+    #
+    def inspect(row_counts)
+      array = [headers.to_csv]
+      if row_counts.nil?
+        return array.join("")
+      end
+      @table.first(row_counts).each do |row|
+        array.push(row.fields.to_csv) unless row.header_row?
+      end
```
kou, line 1045: How about adding `:limit` option to `CSV::Table#to_csv` and use it from `#inspect?`

```ruby
    def to_csv(write_headers: true, limit: nil, **options)
      array = write_headers ? [headers.to_csv(**options)] : []
      limit ||= @table.size
      limit = @table.size + 1 - limit if limit < 0 
      @table.first(limit) do |row|
        array.push(row.fields.to_csv(**options)) unless row.header_row?
      end

      array.join("")
    end

    def inspect
      inspected = +"#<#{self.class} mode:#{@mode} row_count:#{to_a.size}>"
      summary = to_s(limit: 5)
      inspected << "\n" << summary if summary.encoding.ascii_compatible?
      inspected
    end
```

kou (changes requested):
Could you change "fix" in the pull request title?
The current `#inspect` doesn't have a bug. So "fix" isn't suitable for this case.
How about `Improve ...`, `Add contents summary to CSV::Table#inspect` or something?

kou (commented):
It almost looks good.
We can merge this once we add `limit:` documentation to `Table#to_s`.

lib/csv/table.rb
```diff
@@ -999,9 +999,21 @@ def to_a
...
+    #
+    # Defaults option +limit+ to +nil+:
+    #   source = "Name,Value\nfoo,0\nbar,1\nbaz,2\n"
+    #   table = CSV.parse(source, headers: true)
+    #   table.to_csv # => "Name,Value\nfoo,0\nbar,1\nbaz,2\n"
+    #
+    # Limit rows if option +limit+ is given like +2+
+    #   source = "Name,Value\nfoo,0\nbar,1\nbaz,2\n"
+    #   table = CSV.parse(source, limit: 2, headers: true)
+    #   table.to_csv # => "Name,Value\nfoo,0\nbar,1\n"
```
kou, line 1011: Is this example correct?

```suggestion
    #   table = CSV.parse(source, headers: true)
    #   table.to_csv(limit: 2) # => "Name,Value\nfoo,0\nbar,1\n"
```

lib/csv/table.rb
```diff
@@ -999,9 +999,21 @@ def to_a
     # Omits the headers if option +write_headers+ is given as +false+
     # (see {Option +write_headers+}[../CSV.html#class-CSV-label-Option+write_headers]):
     #   table.to_csv(write_headers: false) # => "foo,0\nbar,1\nbaz,2\n"
-    def to_csv(write_headers: true, **options)
+    #
+    # Defaults option +limit+ to +nil+:
+    #   source = "Name,Value\nfoo,0\nbar,1\nbaz,2\n"
+    #   table = CSV.parse(source, headers: true)
+    #   table.to_csv # => "Name,Value\nfoo,0\nbar,1\nbaz,2\n"
```
kou, line 1006: It seems that we can remove this because this is the same code as the above code for `write_headers: true`.

lib/csv/table.rb
ericgpks: I made a mistake, fixed it.

lib/csv/table.rb
ericgpks: I removed these comment.

lib/csv/table.rb
```diff
@@ -999,9 +999,16 @@ def to_a
     # Omits the headers if option +write_headers+ is given as +false+
     # (see {Option +write_headers+}[../CSV.html#class-CSV-label-Option+write_headers]):
     #   table.to_csv(write_headers: false) # => "foo,0\nbar,1\nbaz,2\n"
-    def to_csv(write_headers: true, **options)
+    #
+    # Limit rows if option +limit+ is given like +2+
+    #   source = "Name,Value\nfoo,0\nbar,1\nbaz,2\n"
+    #   table = CSV.parse(source, headers: true)
```
kou, line 1005:
```suggestion
    # Limit rows if option +limit+ is given like +2+:
```

kou:
Thanks!

PR #231, opened 2022-01-13
Improve error reporting for liberal parsing with quoted values

NikolayRys:
In cases when `liberal_parsing` is used with quoted values in a file that has mixed up EOL terminators, this lib produces a meaningless error: 
```
# Normal parsing, no quotes (makes sense)
CSV.parse("A\nB\r\n")
=> CSV::MalformedCSVError: Unquoted fields do not allow new line <"\r\n"> in line 2.
```
```
# Normal parsing, with quotes (kinda makes sense - not super-helpful, but technically correct)
CSV.parse("A\n\"B\"\r\n")
CSV::MalformedCSVError: Any value after quoted field isn't allowed in line 2.
```
```
# Liberal parsing, no quotes ( makes sense)
CSV.parse("A\nB\r\n", liberal_parsing: true)
CSV::MalformedCSVError: Unquoted fields do not allow new line <"\r\n"> in line 2.
```
```
# Liberal parsing, with quotes (misleading)
CSV.parse("A\n\"B\"\r\n", liberal_parsing: true)
CSV::MalformedCSVError: Any value after quoted field isn't allowed in line 2.
```
The last one is highly misleading because liberal parsing does indeed allow values after a quoted field, that is its whole purpose. In addition, there is no information about the actual source of the problem - the wrong newline sequence. 
I personally wasted a lot of time debugging this issue, when received a file with mixed up newline sequences.

This PR adds a separate error message for this case

### DIFF
--- a/lib/csv/parser.rb
+++ b/lib/csv/parser.rb
@@ -229,7 +229,7 @@ class CSV
 
       def keep_drop
         _, _, buffer = @keeps.pop
-        # trace(__method__, :done, :emtpy) unless buffer
+        # trace(__method__, :done, :empty) unless buffer
         return unless buffer
 
         last_keep = @keeps.last
@@ -246,6 +246,10 @@ class CSV
 
       def rest
         @scanner.rest
+      end
+
+      def check(pattern)
+        @scanner.check(pattern)
       end
 
       private
@@ -1022,10 +1026,10 @@ class CSV
           break
         else
           if @quoted_column_value
-            message = if liberal_parsing? and (new_line = @scanner.scan(@line_end))
-              "Illegal end-of-line sequence outside of a quoted field " + "<#{new_line.inspect}>"
+            if liberal_parsing? and (new_line = @scanner.check(@line_end))
+              message = "Illegal end-of-line sequence outside of a quoted field " + "<#{new_line.inspect}>"
             else
-              "Any value after quoted field isn't allowed"
+              message = "Any value after quoted field isn't allowed"
             end
             ignore_broken_line
             raise MalformedCSVError.new(message, @lineno)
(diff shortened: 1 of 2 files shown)


NikolayRys:
I will check the failing tests soon

NikolayRys:
Please re-run the tests

test/csv/parse/test_liberal_parsing.rb
```diff
@@ -28,6 +28,15 @@ def test_middle_quote_end
                  CSV.parse_line(input, liberal_parsing: true))
   end
 
+  def test_endline_after_quoted_field_end
+    input = "A\r\n\"B\"\n"
+    error = assert_raise(CSV::MalformedCSVError) do
+      CSV.parse(input, liberal_parsing: true)
```
kou, line 34: Could you write not-reused value literally?

```suggestion
    error = assert_raise(CSV::MalformedCSVError) do
      CSV.parse("A\r\n\"B\"\n", liberal_parsing: true)
```

lib/csv/parser.rb
```diff
@@ -1022,8 +1022,12 @@ def parse_quotable_robust(&block)
           break
         else
           if @quoted_column_value
+            message = if liberal_parsing? and (new_line = @scanner.scan(@line_end))
```
kou, line 1025: This breaks the following case:

```ruby
require "csv"

csv = CSV.new("A\r\n\"B\"\nC\r\n", liberal_parsing: true)
p csv.gets # => ["A"]
begin
  p csv.gets
rescue
  p :raise
end
p csv.gets # => ["C"]
```

NikolayRys: Thx, will check it out shortly

NikolayRys:
The feedback has been addressed 👍

kou:
Thanks!

Issue #237, opened 2022-02-25
CSV::Table#headers returns [:"", :""] when  header is Japanese


taki:
### Steps to reproduce

```ruby
# frozen_string_literal: true

require 'csv'

header = %w[都道府県 桃の生産量]
rows = [['山梨県', '37,600トン'], ['福島県', '22,500トン'], ['長野県', '12,100トン']]

CSV.open('momo.csv', 'w') do |csv|
  csv << header
  rows.map { |row| csv << row }
end

csv_table = CSV.table('momo.csv')

pp csv_table.headers
```

### Expected behavior

`csv_table#headers` returns `[:都道府県, :桃の生産量]`

### Actual behavior

`csv_table#headers` returns `[:"", :""]`

### System configuration
**Ruby version**: ruby 3.1.0p0
**csv version**: csv (default: 3.2.2)

kou:
I understand your expected behavior but we can't change the current behavior to keep compatibility.

Could you specify `header_converters:` explicitly?

```ruby
csv_table = CSV.table('momo.csv', header_converters: lambda {|h| h.to_sym})
```

kou:
We can provide `:symbol_raw` default header converter by adding a converter to `CSV::HeaderConverters`. If we provide `:symbol_raw`, we can write like the following:

```ruby
csv_table = CSV.table('momo.csv', header_converters: :symbol_raw)
```

@ericgpks Do you want to work on this?

ericgpks:
@kou 
I would like to try to resolve this one! Thanks a lot!!

kou:
@ericgpks Sure. Go ahead!

kou:
Done by #239.

taki:
Thank you!!

PR #239, opened 2022-03-22
feat: add header_converter additional option

ericgpks:
#237 
Add option to convert header written with strings without return empty.

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -928,7 +928,8 @@ class CSV
     symbol:   lambda { |h|
       h.encode(ConverterEncoding).downcase.gsub(/[^\s\w]+/, "").strip.
                                            gsub(/\s+/, "_").to_sym
-    }
+    },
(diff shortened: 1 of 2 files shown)


lib/csv.rb
```diff
@@ -928,7 +928,8 @@ def initialize(message, line_number)
     symbol:   lambda { |h|
       h.encode(ConverterEncoding).downcase.gsub(/[^\s\w]+/, "").strip.
                                            gsub(/\s+/, "_").to_sym
-    }
+    },
+    symbol_raw: lambda { |h| h.to_sym }
```
kou, line 932: Could you convert it to `ConverterEncoding` like others?

```suggestion
    symbol_raw: lambda { |h| h.encode(ConverterEncoding).to_sym }
```

ericgpks: I understood it. Thanks.

kou:
Thanks!

PR #240, opened 2022-03-23
feat: add .idea on gitignore

ericgpks:
Change .gitignore file to ignore .idea and .idea/* files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@
 /pkg/
 /spec/reports/
 /tmp/
+.idea
+.idea/*
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,4 @@
 /pkg/
 /spec/reports/
 /tmp/
-.idea
-.idea/*
+/.idea/


.gitignore
```diff
@@ -5,3 +5,5 @@
 /pkg/
 /spec/reports/
 /tmp/
+.idea
+.idea/*
```
kou, line 9: Does this work?

```suggestion
/.idea/
```

ericgpks: This works enough! Thanks

.gitignore
```diff
@@ -5,3 +5,4 @@
 /pkg/
 /spec/reports/
 /tmp/
+/.idea/
```
kou, line 8: Could you keep this list in alphabetical order?

ericgpks: In this pattern, it is after html or after .bundle ?

kou: `/.bundle/`:

```ruby
pp %w(
/.bundle/
/Gemfile.lock
/coverage/
/html/
/pkg/
/spec/reports/
/tmp/
/.idea/
).sort
```

```text
["/.bundle/",
 "/.idea/",
 "/Gemfile.lock",
 "/coverage/",
 "/html/",
 "/pkg/",
 "/spec/reports/",
 "/tmp/"]
```

amatsuda:
Please allow me to leave a general comment (I'm not a maintainer of this repo, but I just came across this PR).

Basically, a software project's `.gitignore` should not include any user-specific, system-specific, or some other software-specific files.
If an editor or IDE that you use creates some garbage inside your working directory, that file should be ignored via your local `.ignore` mechanism, such as `.config/git/ignore` file.
Otherwise, we have to add `*.swp` and `.idea` and `.vscode` and `*.tmproj` and `nbproject` and `.DS_Store` and a lot more other 3rd party software specific entries to hundreds of millions of Git repositories.

Just for reference, here's how we treat this FAQ on a project that I'm involved in.

https://github.com/rails/rails/pull/7537
https://github.com/rails/rails/pull/24538
https://github.com/rails/rails/pull/42737
https://github.com/rails/rails/pull/541

ericgpks:
Thank you for your comment. I agree with that opinion. So I close this pr.

PR #248, opened 2022-06-15
Add support for RFC 3339 style time

### LINKED ISSUE #247: :date_time converter fails to recognize "YYYY-MM-DD HH:MM"
```ruby
require "csv"
csv = '2022-06-14 18:48'
p CSV.parse(csv, converters: :date_time).first.first #=> "2022-06-14 18:48"
p DateTime.parse(csv) #=> #<DateTime: 2022-06-14T18:48:00+00:00 ((2459745j,67680s,0n),+0s,2299161j)>
```
There are two ways to solve this, I'm not sure which is the best one:
* make the seconds optional:
  ```ruby
  DateTimeMatcher =
    / \A(?: (\w+,?\s+)?\w+\s+\d{1,2}\s+\d{1,2}:\d{1,2}:\d{1,2},?\s+\d{2,4} |
            \d{4}-\d{2}-\d{2}\s\d{2}:\d{2}(?::\d{2})? |
            # ISO-8601
            \d{4}-\d{2}-\d{2}
              (?:T\d{2}:\d{2}(?::\d{2}(?:\.\d+)?(?:[+-]\d{2}(?::\d{2})|Z)?)?)?
        )\z /x
  ```
* or allow "T" to be replaced by " ":
  ```ruby
  DateTimeMatcher =
    / \A(?: (\w+,?\s+)?\w+\s+\d{1,2}\s+\d{1,2}:\d{1,2}:\d{1,2},?\s+\d{2,4} |
            \d{4}-\d{2}-\d{2}\s\d{2}:\d{2}:\d{2} |
            # ISO-8601
            \d{4}-\d{2}-\d{2}
              (?:[T ]\d{2}:\d{2}(?::\d{2}(?:\.\d+)?(?:[+-]\d{2}(?::\d{2})|Z)?)?)?
        )\z /x
  ```

thyresias:
fix #247

ISO 8601 uses "T" for separator of date and time.
RFC 3339 also accepts a space for separator of date and time.

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -875,9 +875,9 @@ class CSV
   # A Regexp used to find and convert some common DateTime formats.
   DateTimeMatcher =
     / \A(?: (\w+,?\s+)?\w+\s+\d{1,2}\s+\d{1,2}:\d{1,2}:\d{1,2},?\s+\d{2,4} |
-            # ISO-8601 and near-ISO (space instead of T) recognized by DateTime.parse
+            # ISO-8601 and RFC-3339 (space instead of T) recognized by DateTime.parse
             \d{4}-\d{2}-\d{2}
-              (?:[T ]\d{2}:\d{2}(?::\d{2}(?:\.\d+)?(?:[+-]\d{2}(?::\d{2})|Z)?)?)?
+              (?:[T\s]\d{2}:\d{2}(?::\d{2}(?:\.\d+)?(?:[+-]\d{2}(?::\d{2})|Z)?)?)?
         )\z /x
 
   # The encoding used by all converters.
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -875,10 +875,9 @@ class CSV
   # A Regexp used to find and convert some common DateTime formats.
   DateTimeMatcher =
     / \A(?: (\w+,?\s+)?\w+\s+\d{1,2}\s+\d{1,2}:\d{1,2}:\d{1,2},?\s+\d{2,4} |
-            \d{4}-\d{2}-\d{2}\s\d{2}:\d{2}:\d{2} |
-            # ISO-8601
+            # ISO-8601 and near-ISO (space instead of T) recognized by DateTime.parse
             \d{4}-\d{2}-\d{2}
-              (?:T\d{2}:\d{2}(?::\d{2}(?:\.\d+)?(?:[+-]\d{2}(?::\d{2})|Z)?)?)?
+              (?:[T ]\d{2}:\d{2}(?::\d{2}(?:\.\d+)?(?:[+-]\d{2}(?::\d{2})|Z)?)?)?
         )\z /x
 
   # The encoding used by all converters.
--- a/test/csv/test_data_converters.rb
+++ b/test/csv/test_data_converters.rb
@@ -187,5 +187,4 @@ class TestCSVDataConverters < Test::Unit::TestCase
     assert_equal(datetime,
                  CSV::Converters[:date_time][rfc3339_string])
   end
-
 end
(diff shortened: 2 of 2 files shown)


kou:
@ericgpks Do you want to review this?

ericgpks:
I would like to try review.

lib/csv.rb
```diff
@@ -875,10 +875,9 @@ def initialize(message, line_number)
   # A Regexp used to find and convert some common DateTime formats.
   DateTimeMatcher =
     / \A(?: (\w+,?\s+)?\w+\s+\d{1,2}\s+\d{1,2}:\d{1,2}:\d{1,2},?\s+\d{2,4} |
-            \d{4}-\d{2}-\d{2}\s\d{2}:\d{2}:\d{2} |
-            # ISO-8601
+            # ISO-8601 and near-ISO (space instead of T) recognized by DateTime.parse
             \d{4}-\d{2}-\d{2}
-              (?:T\d{2}:\d{2}(?::\d{2}(?:\.\d+)?(?:[+-]\d{2}(?::\d{2})|Z)?)?)?
+              (?:[T ]\d{2}:\d{2}(?::\d{2}(?:\.\d+)?(?:[+-]\d{2}(?::\d{2})|Z)?)?)?
```
ericgpks, line 880: I don't know if this meets a condition mentioned in [this comment](https://github.com/ruby/csv/issues/247#issuecomment-1155662500)

ericgpks: @kou 
Can you help me?

kou: @thyresias It seems that this approach is the latter approach not the former approach. Is it intentional?

Anyway, could you refer [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) instead of "near-ISO" for "YYYY-MM-DD hh:mm:ss" format?

This rejects `"YYYY-MM-DD\thh:mm:dd"` that is accepted without this change. Could you use `[T\s]` instead of `[T ]`?

test/csv/test_data_converters.rb
```diff
@@ -103,4 +103,19 @@ def test_builtin_date_time_converter_iso8601_utc
     assert_equal(datetime,
                  CSV::Converters[:date_time][iso8601_string])
   end
+
+  def test_builtin_date_time_converter_near_iso
```
kou, line 107: Could you split this test to multiple tests like existing `test_bulitin_date_time_converter_iso8601_*` for easy to debug on failure?
If `"2018-01-14 22:25"` is failed, the rest cases such as `"2018-01-14 22:25:19"` aren't executed by this style. The results of all cases are helpful to debug.

test/csv/test_data_converters.rb
```diff
@@ -103,4 +103,89 @@ def test_builtin_date_time_converter_iso8601_utc
     assert_equal(datetime,
                  CSV::Converters[:date_time][iso8601_string])
   end
+
+  def test_builtin_date_time_converter_rfc3339_minute
+    iso8601_string = "2018-01-14 22:25"
```
kou, line 108: Could you use `rfc3339_string` instead of `iso8601_string` because this is not ISO 8601 format?

thyresias: done.

test/csv/test_data_converters.rb
```diff
@@ -103,4 +103,89 @@ def test_builtin_date_time_converter_iso8601_utc
...
+                 CSV::Converters[:date_time][rfc3339_string])
+  end
+
+  def test_builtin_date_time_converter_rfc3339_tab_utc
+    rfc3339_string = "2018-01-14\t22:25:19Z"
+    datetime = DateTime.new(2018, 1, 14, 22, 25, 19)
+    assert_equal(datetime,
+                 CSV::Converters[:date_time][rfc3339_string])
+  end
+
```
kou, line 190:
```suggestion
```

kou:
Thanks!

PR #249, opened 2022-06-17
Simplify odd division by two in Parser#parse_quoted_column_value

Maumagnaguagno:
Division by two of odd ``n_quotes`` (integer >= 0) does not need to decrement one to generate the same result as even numbers.
Division by two already removes the least significant bit that is being subtracted: (0bxxx1 - 1) / 2 => 0bxxx0 / 2 => 0bxxx.

### DIFF
--- a/lib/csv/parser.rb
+++ b/lib/csv/parser.rb
@@ -1167,7 +1167,7 @@ class CSV
             break
           else
             value << quotes[0, n_quotes / 2]
-            break if n_quotes % 2 == 1
+            break if (n_quotes % 2) == 1
           end
         end
         value
--- a/lib/csv/parser.rb
+++ b/lib/csv/parser.rb
@@ -1141,7 +1141,7 @@ class CSV
       if (n_quotes % 2).zero?
         quotes[0, (n_quotes - 2) / 2]
       else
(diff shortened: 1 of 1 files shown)


olleolleolle:
@Maumagnaguagno They may give the same result, but my guess is that the one operating on even numbers works quicker.

Maumagnaguagno:
Yes, the formula operating on even numbers is "quicker", ``n_quotes / 2``, than the one for odds, ``(n_quotes - 1) / 2``, which is one of the reasons I want to use the same formula for both even and odds. The main reason is to simplify the code and make clearer that whats is pushed to ``value`` is the same for both cases.

lib/csv/parser.rb
```diff
@@ -1165,11 +1165,9 @@ def parse_quoted_column_value
           n_quotes = quotes.size
           if n_quotes == 1
             break
-          elsif (n_quotes % 2) == 1
-            value << quotes[0, (n_quotes - 1) / 2]
-            break
           else
             value << quotes[0, n_quotes / 2]
+            break if n_quotes % 2 == 1
```
kou, line 1170:
```suggestion
            break if (n_quotes % 2) == 1
```

kou:
Thanks!

PR #250, opened 2022-06-18
Remove no longer required refinements

Maumagnaguagno:
Since PR #159, the minimum Ruby version is 2.5.0, a version which no longer requires refinements for String#delete_suffix?, String#match? and Regexp#match?.


kou:
Good catch!

PR #251, opened 2022-06-23
Readable prefix/suffix removal in Parser#strip_value

Maumagnaguagno:
The use of ``String#delete_prefix!/delete_suffix!`` makes this method more readable.
It also makes possible to strip strings with more than one character.

Note that I am using the methods ending with ``!`` to modify ``value``, this matches the already used ``strip!``.
I also modified the early return to show that ``value`` is always returned, and with this PR all ``value`` modifications happen in-place.
This could be used to further simplify the method itself (no need to return a value) and the single usage I have found in ``Parser#parse_no_quote``: ``line = strip_value(line)`` to ``strip_value(line)``.

I have not done this modification yet to leave it open for discussion.

### DIFF
--- a/lib/csv/parser.rb
+++ b/lib/csv/parser.rb
@@ -1197,12 +1197,17 @@ class CSV
     end
 
     def strip_value(value)
-      return value if not @strip or value.nil?
+      return value unless @strip
+      return value if value.nil?
 
       case @strip
       when String
-        nil while value.delete_prefix!(@strip)
-        nil while value.delete_suffix!(@strip)
+        while value.delete_prefix!(@strip)
+          # do nothing
+        end
+        while value.delete_suffix!(@strip)
+          # do nothing
+        end
       else
         value.strip!
       end
--- a/lib/csv/parser.rb
+++ b/lib/csv/parser.rb
@@ -1197,20 +1197,12 @@ class CSV
     end
 
     def strip_value(value)
-      return value unless @strip
-      return nil if value.nil?
+      return value if not @strip or value.nil?
 
       case @strip
       when String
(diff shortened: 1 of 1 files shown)


lib/csv/parser.rb
```diff
@@ -1197,20 +1197,12 @@ def parse_row_end
     end
 
     def strip_value(value)
-      return value unless @strip
-      return nil if value.nil?
+      return value if not @strip or value.nil?
```
kou, line 1200: I don't like this unification because of readability.
If you want to show `value` is always returns explicitly, how about just changing to `return value if value.nil?`?

lib/csv/parser.rb
```diff
@@ -1197,20 +1197,12 @@ def parse_row_end
...
-        size = value.size
-        while value.start_with?(@strip)
-          size -= 1
-          value = value[1, size]
-        end
-        while value.end_with?(@strip)
-          size -= 1
-          value = value[0, size]
-        end
+        nil while value.delete_prefix!(@strip)
```
kou, line 1204: I don't think that this is readable.
How about the following?

```suggestion
        while value.delete_prefix!(@strip)
          # do nothing
        end
```

lib/csv/parser.rb
Maumagnaguagno: Sure, I will modify this to keep both cases separate.

lib/csv/parser.rb
Maumagnaguagno: I just did not want to have an empty while block. I will do the suggested changes.

kou:
Thanks!

Issue #252, opened 2022-06-30
Add quoted information to CSV::FieldInfo


kou:
Some CSV writers use quote with special meaning. For example, R's `write.csv` use quote for string value or factor value.

 https://www.rdocumentation.org/packages/utils/versions/3.6.2/topics/write.table
 
 > quote
>
> a logical value (TRUE or FALSE) or a numeric vector. If TRUE, any character or factor columns will be surrounded by double quotes. If a numeric vector, its elements are taken as the indices of columns to quote. In both cases, row and column names are quoted if they are written. If FALSE, nothing is quoted.

The current CSV parser doesn't provide quoted information to field converter. If field convert knows whether the target value is quoted or not, it can process the target value as a factor value. 

See also: https://github.com/red-data-tools/red-datasets/pull/140#issuecomment-1164855481

heronshoes:
:+1:

For example,

```ruby
str = <<~CSV
  "serial","value"
  "108",11
  "109",12
  "10A",13
  "10B",14
CSV
CSV.parse(str, converters: [:numeric])

# =>
[["serial", "value"], [108, 11], [109, 12], ["10A", 13], ["10B", 14]]
```
I want to preserve quoted numeric values "108", "109", but there are no means to do this in converter.

PR #253, opened 2022-07-04
Add quoted information to CSV::FieldInfo #252

heronshoes:
This is a proposal to fix the issue #252;
- Add a new member `:quoted?` to `FieldInfo` .
- ~~Introduce an array of booleans `@quoted_fields` in Parser to collect original column value was quoted?.~~
- Check in each of `parse_no_quote`, `parse_quotable_loose` and `parse_quotable_robust`.
  - Use local variable `quoted_fields` to collect booleans in each method and pass it through `#emit_row`.
- Add test of custom converter for preserving quoted value.

### DIFF
--- a/lib/csv/fields_converter.rb
+++ b/lib/csv/fields_converter.rb
@@ -44,7 +44,7 @@ class CSV
       @converters.empty?
     end
 
-    def convert(fields, headers, lineno)
+    def convert(fields, headers, lineno, quoted_fields)
       return fields unless need_convert?
 
       fields.collect.with_index do |field, index|
@@ -63,7 +63,8 @@ class CSV
             else
               header = nil
             end
-            field = converter[field, FieldInfo.new(index, lineno, header)]
+            quoted = quoted_fields[index]
+            field = converter[field, FieldInfo.new(index, lineno, header, quoted)]
           end
           break unless field.is_a?(String)  # short-circuit pipeline for speed
         end
--- a/lib/csv/parser.rb
+++ b/lib/csv/parser.rb
@@ -770,7 +770,7 @@ class CSV
         @use_headers = true
       end
       if @raw_headers
-        @headers = adjust_headers(@raw_headers)
+        @headers = adjust_headers(@raw_headers, [])
       else
         @headers = nil
       end
--- a/lib/csv/parser.rb
+++ b/lib/csv/parser.rb
@@ -783,8 +783,9 @@ class CSV
                      quote_char: @quote_character)
     end
 
-    def adjust_headers(headers)
-      adjusted_headers = @header_fields_converter.convert(headers, nil, @lineno)
+    def adjust_headers(headers, quoted_fields = nil)
+      quoted_fields = [] unless quoted_fields
+      adjusted_headers = @header_fields_converter.convert(headers, nil, @lineno, quoted_fields)
       adjusted_headers.each {|h| h.freeze if h.is_a? String}
       adjusted_headers
     end
@@ -928,9 +929,11 @@ class CSV
         if line.empty?
           next if @skip_blanks
           row = []
+          quoted_fields = []
         else
           line = strip_value(line)
           row = line.split(@split_column_separator, -1)
+          quoted_fields = []
           if @max_field_size
             row.each do |column|
               validate_field_size(column)
@@ -944,7 +947,7 @@ class CSV
           end
         end
         @last_line = original_line
-        emit_row(row, &block)
+        emit_row(row, quoted_fields, &block)
       end
     end
 
@@ -966,25 +969,30 @@ class CSV
             next
           end
           row = []
+          quoted_fields = []
         elsif line.include?(@cr) or line.include?(@lf)
           @scanner.keep_back
           @need_robust_parsing = true
           return parse_quotable_robust(&block)
         else
           row = line.split(@split_column_separator, -1)
+          quoted_fields = []
           n_columns = row.size
           i = 0
           while i < n_columns
             column = row[i]
             if column.empty?
+              quoted_fields << false
               row[i] = nil
             else
               n_quotes = column.count(@quote_character)
               if n_quotes.zero?
+                quoted_fields << false
                 # no quote
               elsif n_quotes == 2 and
                    column.start_with?(@quote_character) and
                    column.end_with?(@quote_character)
+                quoted_fields << true
                 row[i] = column[1..-2]
               else
                 @scanner.keep_back
@@ -999,13 +1007,14 @@ class CSV
         @scanner.keep_drop
         @scanner.keep_start
         @last_line = original_line
-        emit_row(row, &block)
+        emit_row(row, quoted_fields, &block)
       end
       @scanner.keep_drop
     end
 
     def parse_quotable_robust(&block)
       row = []
+      quoted_fields = []
       skip_needless_lines
       start_row
       while true
@@ -1021,18 +1030,19 @@ class CSV
           row << value
         elsif parse_row_end
           if row.empty? and value.nil?
-            emit_row([], &block) unless @skip_blanks
+            emit_row([], [], &block) unless @skip_blanks
           else
             row << value
-            emit_row(row, &block)
+            emit_row(row, quoted_fields, &block)
             row = []
+            quoted_fields = []
           end
           skip_needless_lines
           start_row
         elsif @scanner.eos?
           break if row.empty? and value.nil?
           row << value
-          emit_row(row, &block)
+          emit_row(row, quoted_fields, &block)
           break
         else
           if @quoted_column_value
@@ -1229,22 +1239,22 @@ class CSV
       @scanner.keep_start
     end
 
-    def emit_row(row, &block)
+    def emit_row(row, quoted_fields, &block)
       @lineno += 1
 
       raw_row = row
       if @use_headers
         if @headers.nil?
-          @headers = adjust_headers(row)
+          @headers = adjust_headers(row, quoted_fields)
           return unless @return_headers
           row = Row.new(@headers, row, true)
         else
           row = Row.new(@headers,
-                        @fields_converter.convert(raw_row, @headers, @lineno))
+                        @fields_converter.convert(raw_row, @headers, @lineno, quoted_fields))
         end
       else
         # convert fields, if needed...
-        row = @fields_converter.convert(raw_row, nil, @lineno)
+        row = @fields_converter.convert(raw_row, nil, @lineno, quoted_fields)
       end
 
       # inject unconverted fields and accessor, if requested...
--- a/lib/csv/parser.rb
+++ b/lib/csv/parser.rb
@@ -758,9 +758,10 @@ class CSV
       case headers
       when Array
         @raw_headers = headers
+        quoted_fields = [false] * @raw_headers.size
         @use_headers = true
       when String
-        @raw_headers = parse_headers(headers)
+        @raw_headers, quoted_fields = parse_headers(headers)
         @use_headers = true
       when nil, false
         @raw_headers = nil
@@ -770,21 +771,27 @@ class CSV
         @use_headers = true
       end
       if @raw_headers
-        @headers = adjust_headers(@raw_headers, [])
+        @headers = adjust_headers(@raw_headers, quoted_fields)
       else
         @headers = nil
       end
     end
 
     def parse_headers(row)
-      CSV.parse_line(row,
-                     col_sep:    @column_separator,
-                     row_sep:    @row_separator,
-                     quote_char: @quote_character)
+      quoted_fields = []
+      converter = lambda do |field, info|
+        quoted_fields << info.quoted?
+        field
+      end
+      headers = CSV.parse_line(row,
+                               col_sep:    @column_separator,
+                               row_sep:    @row_separator,
+                               quote_char: @quote_character,
+                               converters: [converter])
+      [headers, quoted_fields]
     end
 
     def adjust_headers(headers, quoted_fields)
-      quoted_fields = [] unless quoted_fields
       adjusted_headers = @header_fields_converter.convert(headers, nil, @lineno, quoted_fields)
       adjusted_headers.each {|h| h.freeze if h.is_a? String}
       adjusted_headers
@@ -933,7 +940,7 @@ class CSV
         else
           line = strip_value(line)
           row = line.split(@split_column_separator, -1)
-          quoted_fields = []
+          quoted_fields = [false] * row.size
           if @max_field_size
             row.each do |column|
               validate_field_size(column)
--- a/lib/csv/parser.rb
+++ b/lib/csv/parser.rb
@@ -783,7 +783,7 @@ class CSV
                      quote_char: @quote_character)
     end
 
-    def adjust_headers(headers, quoted_fields = nil)
+    def adjust_headers(headers, quoted_fields)
       quoted_fields = [] unless quoted_fields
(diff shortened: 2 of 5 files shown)


lib/csv/fields_converter.rb
```diff
@@ -44,7 +44,7 @@ def empty?
       @converters.empty?
     end
 
-    def convert(fields, headers, lineno)
+    def convert(fields, headers, lineno, quoted_fields = nil)
```
kou, line 47: Can we make `quoted_fields` mandatory argument? If we use `nil` here, `FieldInfo#quoted?` returns `nil` not boolean.

lib/csv/parser.rb
```diff
@@ -928,9 +928,11 @@ def parse_no_quote(&block)
         if line.empty?
           next if @skip_blanks
           row = []
+          @quoted_fields = []
```
kou, line 931: How about passing `quoted_fields` to `#emit_row` and make this a local variable?

lib/csv/fields_converter.rb
heronshoes: Thanks. I thought it is not required when converting headers.
But it is better to explicitly pass an argument for them.

lib/csv/parser.rb
heronshoes: Thanks.
This is simpler and localized !

lib/csv/fields_converter.rb
kou: Headers may be parsed from string. So quoted information for headers may be useful.

heronshoes: I understood, thank you. So currently no information is passed in `#adjust_headers`  ....

heronshoes: I added support for quoted header converters.

heronshoes:
I rebased feat/issue-252 to the latest HEAD.

lib/csv/parser.rb
```diff
@@ -783,8 +783,9 @@ def parse_headers(row)
                      quote_char: @quote_character)
     end
 
-    def adjust_headers(headers)
-      adjusted_headers = @header_fields_converter.convert(headers, nil, @lineno)
+    def adjust_headers(headers, quoted_fields = nil)
```
kou, line 786: Can we always require `quoted_fields` like `emit_raw`?

```suggestion
    def adjust_headers(headers, quoted_fields)
```

lib/csv/parser.rb
```diff
@@ -1021,18 +1030,19 @@ def parse_quotable_robust(&block)
           row << value
```
kou, line 1030: Do we need add an element to `quoted_fields` here?

```suggestion
          row << value
          quoted_fields << @quoted_column_value
```

Could you also add a test that uses this method? `CSV.parse(..., strip: true)` will use this method.

lib/csv/parser.rb
```diff
@@ -1021,18 +1030,19 @@ def parse_quotable_robust(&block)
...
         elsif parse_row_end
           if row.empty? and value.nil?
-            emit_row([], &block) unless @skip_blanks
+            emit_row([], [], &block) unless @skip_blanks
           else
             row << value
```
kou, line 1035: Do we need add an element to `quoted_field`s here?

```suggestion
            row << value
            quoted_fields << @quoted_column_value
```

lib/csv/parser.rb
```diff
@@ -1021,18 +1030,19 @@ def parse_quotable_robust(&block)
...
-            emit_row(row, &block)
+            emit_row(row, quoted_fields, &block)
             row = []
+            quoted_fields = []
           end
           skip_needless_lines
           start_row
         elsif @scanner.eos?
           break if row.empty? and value.nil?
           row << value
```
kou, line 1044: Do we need add an element to `quoted_field`s here?

```suggestion
          row << value
          quoted_fields << @quoted_column_value
```

lib/csv/writer.rb
```diff
@@ -39,7 +39,7 @@ def <<(row)
       @headers ||= row if @use_headers
       @lineno += 1
 
-      row = @fields_converter.convert(row, nil, lineno) if @fields_converter
+      row = @fields_converter.convert(row, nil, lineno, []) if @fields_converter
```
kou, line 42:
```suggestion
      if @fields_converter
        quoted_fields = [false] * row.size
        row = @fields_converter.convert(row, nil, lineno, quoted_fields)
      end
```

lib/csv/writer.rb
```diff
@@ -94,7 +94,7 @@ def prepare_header
       return unless @headers
 
       converter = @options[:header_fields_converter]
-      @headers = converter.convert(@headers, nil, 0)
+      @headers = converter.convert(@headers, nil, 0, [])
```
kou, line 97: We need to care about a case that `headers` is the following `String`:

```csv
"quoted",unquoted,"quoted-again"
```

Could you also add a test for this case?

test/csv/parse/test_convert.rb
```diff
@@ -107,4 +107,45 @@ def test_empty_value
...
+      @preserving_converter = lambda do |field, info|
+        f = field.encode(CSV::ConverterEncoding)
+        return field if info.quoted?
+        return DateTime.parse(f) if f.match?(DateTimeMatcher) rescue
+        return Integer(f) rescue
+        return Float(f) rescue
+        field
+      end
+
+      @str = <<~CSV
```
kou, line 122:
```suggestion
      @csv = <<~CSV
```

test/csv/parse/test_convert.rb
```diff
@@ -107,4 +107,45 @@ def test_empty_value
     assert_equal([nil, "empty", "a"],
                  CSV.parse_line(',"",a', empty_value: "empty"))
   end
+
+  sub_test_case("add quoted info") do
```
kou, line 111:
```suggestion
  sub_test_case("#quoted?") do
```

test/csv/parse/test_convert.rb
```diff
@@ -107,4 +107,45 @@ def test_empty_value
...
+    def setup
+      @preserving_converter = lambda do |field, info|
+        f = field.encode(CSV::ConverterEncoding)
+        return field if info.quoted?
+        return DateTime.parse(f) if f.match?(DateTimeMatcher) rescue
+        return Integer(f) rescue
+        return Float(f) rescue
+        field
```
kou, line 119: Can we simplify this?

```suggestion
        f = field.encode(CSV::ConverterEncoding)
        return f if info.quoted?
        Integer(f, 10)
```

test/csv/parse/test_convert.rb
```diff
@@ -107,4 +107,45 @@ def test_empty_value
...
+      end
+
+      @str = <<~CSV
+        "serial",value
+        "109",12
+        "10A",13
+      CSV
+    end
+
+    def test_custom_field_info_quoted_parse_line
```
kou, line 129:
```suggestion
    def test_parse_line
```

test/csv/parse/test_convert.rb
```diff
@@ -107,4 +107,45 @@ def test_empty_value
...
+      row = CSV.parse_line('1,"2",3', converters: @preserving_converter)
+      assert_equal([1, "2", 3], row)
+    end
+
+    def test_custom_field_info_quoted_parse
```
kou, line 134:
```suggestion
    def test_parse
```

test/csv/parse/test_convert.rb
```diff
@@ -107,4 +107,45 @@ def test_empty_value
...
+      expected = [["serial", "value"], ["109", 12], ["10A", 13]]
+      row = CSV.parse(@str, converters: @preserving_converter)
```
kou, line 136: How about putting the parsed CSV (`@str`) here instead of creating it in `setup`?
It'll be easy to read what is tested here by aligning the expected value with input value.

test/csv/parse/test_convert.rb
```diff
@@ -107,4 +107,45 @@ def test_empty_value
...
+      assert_equal(expected, row)
+    end
+
+    def test_custom_field_info_quoted_parse_header
```
kou, line 140:
```suggestion
    def test_parse_headers
```

test/csv/parse/test_convert.rb
```diff
@@ -107,4 +107,45 @@ def test_empty_value
...
+      quoted_header_converter = lambda do |field, info|
+        f = field.encode(CSV::ConverterEncoding)
+        return field if info.quoted?
+        field.to_sym
```
kou, line 144:
```suggestion
        f = field.encode(CSV::ConverterEncoding)
        return f if info.quoted?
        f.to_sym
```

test/csv/parse/test_convert.rb
```diff
@@ -107,4 +107,45 @@ def test_empty_value
...
+
+    def test_custom_field_info_quoted_parse_line
+      row = CSV.parse_line('1,"2",3', converters: @preserving_converter)
+      assert_equal([1, "2", 3], row)
+    end
+
+    def test_custom_field_info_quoted_parse
+      expected = [["serial", "value"], ["109", 12], ["10A", 13]]
+      row = CSV.parse(@str, converters: @preserving_converter)
+      assert_equal(expected, row)
```
kou, line 137:
```suggestion
      rows = CSV.parse(@str, converters: @preserving_converter)
      assert_equal(expected, rows)
```

test/csv/parse/test_convert.rb
```diff
@@ -107,4 +107,45 @@ def test_empty_value
...
+
+    def test_custom_field_info_quoted_parse_header
+      quoted_header_converter = lambda do |field, info|
+        f = field.encode(CSV::ConverterEncoding)
+        return field if info.quoted?
+        field.to_sym
+      end
+      expected = [["serial", :value], ["109", "12"], ["10A", "13"]]
+      row = CSV.parse(@str, headers: true, header_converters: quoted_header_converter)
+      assert_equal(expected, row.to_a)
```
kou, line 148:
```suggestion
      table = CSV.parse(@str, headers: true, header_converters: quoted_header_converter)
      assert_equal(expected, table.to_a)
```

test/csv/parse/test_convert.rb
```diff
@@ -107,4 +107,45 @@ def test_empty_value
...
+    end
+
+    def test_custom_field_info_quoted_parse_header
+      quoted_header_converter = lambda do |field, info|
+        f = field.encode(CSV::ConverterEncoding)
+        return field if info.quoted?
+        field.to_sym
+      end
+      expected = [["serial", :value], ["109", "12"], ["10A", "13"]]
+      row = CSV.parse(@str, headers: true, header_converters: quoted_header_converter)
```
kou, line 147: Could you also add `CSV.parse(..., headers: '"serial",value', ...)` (specifying headers as `String`) case?

kou (commented):
Thanks!
(If you want me to write code instead of commenting, please let me know.)

heronshoes:
Thanks! I will check.
(I think this style is better because I can check one by one.)

heronshoes:
I will add test case in:
- "quoted",unquoted,"quoted-again"
- Put expected @str in assertion locally
- CSV.parse(..., headers: '"serial",value', ...) (specifying headers as String) case

Please hold approving.

heronshoes:
I added and fixed except 3rd one:
- CSV.parse(..., headers: '"serial",value', ...) (specifying headers as String) case

This one requires fix in `prepare_header`, `parse_headers` of parser.rb, but I can't fix yet.
It may also require `Symbol#encode` fix.

Current CSV has a bug in converting when specified header has a Symbol.
```ruby
CSV.parse("1,2", headers: ["string", :symbol], header_converters: :symbol)
```

`CSV.parse(..., headers: '"serial",value', ...)` case requires prior parsing, so same issue will happen.

kou:
How about this?

```diff
diff --git a/lib/csv/parser.rb b/lib/csv/parser.rb
index c3c1eff..cd5a45f 100644
--- a/lib/csv/parser.rb
+++ b/lib/csv/parser.rb
@@ -758,9 +758,10 @@ class CSV
       case headers
       when Array
         @raw_headers = headers
+        quoted_fields = [false] * @raw_headers.size
         @use_headers = true
       when String
-        @raw_headers = parse_headers(headers)
+        @raw_headers, quoted_fields = parse_headers(headers)
         @use_headers = true
       when nil, false
         @raw_headers = nil
@@ -770,20 +771,24 @@ class CSV
         @use_headers = true
       end
       if @raw_headers
-        # Set 'quoted?' info to false to avoid re-converting
-        @headers = adjust_headers(@raw_headers, [true] * @raw_headers.size)
+        @headers = adjust_headers(@raw_headers, quoted_fields)
       else
         @headers = nil
       end
     end
 
     def parse_headers(row)
-      converters = @return_headers ? [] : @header_fields_converter.to_a
-      CSV.parse_line(row,
-                     col_sep:    @column_separator,
-                     row_sep:    @row_separator,
-                     quote_char: @quote_character,
-                     converters: converters)
+      quoted_fields = []
+      converter = lambda do |field, info|
+        quoted_fields << info.quoted?
+        field
+      end
+      headers = CSV.parse_line(row,
+                               col_sep:    @column_separator,
+                               row_sep:    @row_separator,
+                               quote_char: @quote_character,
+                               converters: [converter])
+      [headers, quoted_fields]
     end
 
     def adjust_headers(headers, quoted_fields)
```

heronshoes:
Thank you!! I reflected this code!

lib/csv/parser.rb
```diff
@@ -770,21 +771,28 @@ def prepare_header
...
+      quoted_fields = []
+      converter = lambda do |field, info|
+        quoted_fields << info.quoted?
+        field
+      end
+      headers = CSV.parse_line( row,
+                                col_sep:    @column_separator,
+                                row_sep:    @row_separator,
+                                quote_char: @quote_character,
+                                converters: [converter])
```
kou, line 790:
```suggestion
      headers = CSV.parse_line(row,
                               col_sep:    @column_separator,
                               row_sep:    @row_separator,
                               quote_char: @quote_character,
                               converters: [converter])
```

lib/csv/parser.rb
```diff
@@ -928,9 +936,11 @@ def parse_no_quote(&block)
         if line.empty?
           next if @skip_blanks
           row = []
+          quoted_fields = []
         else
           line = strip_value(line)
           row = line.split(@split_column_separator, -1)
+          quoted_fields = []
```
kou, line 943:
```suggestion
          quoted_fields = [false] * row.size
```

test/csv/parse/test_convert.rb
```diff
@@ -107,4 +107,69 @@ def test_empty_value
...
+        "109",1
+        "10A",2
+      CSV
+      assert_equal(expected, table.to_a)
+    end
+  end
+
+  sub_test_case("converter with #encode for Symbols") do
+    def test_encode_symbol
+      assert_nothing_raised { CSV.parse("1,2", headers: ["string", :symbol] , header_converters: :symbol) }
```
kou, line 172: It seems that this is not related to this pull request.
Could you remove this from this pull request? If this is needed, please open another pull request.

kou (commented):
Thanks!
It looks almost good! Could you confirm my comments?

test/csv/parse/test_convert.rb
heronshoes: Thank you. I removed last one and create modified commit.

I will consider more for another pull request about converting when specified header has a Symbol.

kou:
Thanks!

PR #256, opened 2022-08-06
feat: add generate_lines method

ericgpks:


lib/csv.rb
```diff
@@ -1464,6 +1464,46 @@ def generate_line(row, **options)
...
+    #   CSV.generate_lines(rows)
+    #   CSV.generate_lines(rows, **options)
+    #
+    # Returns the \String created by generating \CSV from
+    # using the specified +options+.
+    #
+    # Argument +rows+ must be an \Array of row. Row is \Array of \String or \CSV::Row.
+    #
+    # Special options:
+    # * Option <tt>:row_sep</tt> defaults to <tt>"\n"> on Ruby 3.0 or later
```
kou, line 1477:
```suggestion
    # * Option <tt>:row_sep</tt> defaults to <tt>"\n"</tt> on Ruby 3.0 or later
```

lib/csv.rb
```diff
@@ -1464,6 +1464,46 @@ def generate_line(row, **options)
...
+    #   the first non-+nil+ field in +row+, if possible, but you may need to use
+    #   this parameter as a backup plan.
+    #
+    # For other +options+,
+    # see {Options for Generating}[#class-CSV-label-Options+for+Generating].
+    #
+    # ---
+    #
+    # Returns the \String generated from an
+    #   CSV.generate_lines(['foo', '0'],['bar', '1'],['baz', '2']) # => "foo,0\r\nbar,1\r|nbaz.2"
```
kou, line 1491:
```suggestion
    #   CSV.generate_lines(['foo', '0'], ['bar', '1'], ['baz', '2']) # => "foo,0\r\nbar,1\r\nnbaz,2\r\n"
```

BTW, is `\r\n` really used for row separator? The above document says "Option `:row_sep` defaults to `"\n"`.

test/csv/interface/test_write.rb
```diff
@@ -78,6 +78,15 @@ def test_generate_line_row_sep
     LINE
   end
 
+  def test_generate_lines
```
kou, line 81: Could you move this under `#test_generate_line_shortcut`?
I want to keep `test_generate_line*` grouping.

kou (commented):
Sorry for my late review!

test/csv/interface/test_write.rb
ericgpks: Sure!

lib/csv.rb
ericgpks: In the `general_line` method, it is written like this `CSV.generate_line(['foo', '0']) # => "foo,0\n"` . So I think that I should write the same as this.

kou:
Thanks!

PR #259, opened 2022-09-03
Use memoized parser_enumerator directly instead of assigning to variable

sampatbadhe:
Use memoized parser_enumerator directly instead of assigning to variable

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -2587,10 +2587,9 @@ class CSV
   #   csv.read
   def read
     rows = []
-    enumerator = parser_enumerator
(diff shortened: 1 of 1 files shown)


kou:
Why?

sampatbadhe:
`parser_enumerator` is already memoized, so I feel no need to assign `parser_enumerator` to any variable to use.

kou:
I want to avoid method calls as much as possible here.

sampatbadhe:
sure @kou :+1: 
I'll close this PR.

PR #262, opened 2022-11-10
Remove key sorting in CSV.instance

Maumagnaguagno:
``DEFAULT_OPTIONS`` is a frozen Hash with symbols as keys.
There is no need to convert such keys to strings before sorting.
Note that sorting is not needed for stable implementations of Hash, keys would always be returned in the same entry order.
Sorting removed as all supported Ruby versions implement entry order Hashes.

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1005,7 +1005,7 @@ class CSV
     def instance(data = $stdout, **options)
       # create a _signature_ for this method call, data object and options
       sig = [data.object_id] +
-            options.values_at(*DEFAULT_OPTIONS.keys.sort_by { |sym| sym.to_s })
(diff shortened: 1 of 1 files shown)


lib/csv.rb
```diff
@@ -1005,7 +1005,7 @@ class << self
     def instance(data = $stdout, **options)
       # create a _signature_ for this method call, data object and options
       sig = [data.object_id] +
-            options.values_at(*DEFAULT_OPTIONS.keys.sort_by { |sym| sym.to_s })
+            options.values_at(*DEFAULT_OPTIONS.keys.sort!)
```
kou, line 1008: We can remove sorting here because we don't support old Ruby that doesn't use inserted-order for `Hash`:

```suggestion
            options.values_at(*DEFAULT_OPTIONS.keys)
```

PR #278, opened 2023-04-03
test: rename Helper to CSVHelper

luke-gru:
Rename it so that in ruby/ruby, the generic name Helper is not used.

### DIFF
--- a/test/csv/helper.rb
+++ b/test/csv/helper.rb
@@ -5,7 +5,7 @@ require "csv"
 
 require_relative "../lib/with_different_ofs"
 
-module Helper
+module CSVHelper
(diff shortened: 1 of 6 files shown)


luke-gru:
I'm assuming changes to this repo are ported over to ruby/ruby, right?

kou:
> I'm assuming changes to this repo are ported over to ruby/ruby, right?

Right. But it's a manual operation.
Do you want this in ruby/ruby right now?

luke-gru:
No, take your time there's no rush.

kou:
I see.
I'll sync it to ruby/ruby when we release a new version or the next Ruby preview release.

PR #283, opened 2023-06-28
[DOC] Fix broken links

BurdetteLamar:


lib/csv.rb
```diff
@@ -106,7 +106,7 @@
 #
 # If you are familiar with \CSV data and have a particular task in mind,
 # you may want to go directly to the:
-# - {Recipes for CSV}[doc/csv/recipes/recipes_rdoc.html].
+# - {Recipes for CSV}[https://doc.rubylang.org/doc/csv/recipes/recipes_rdoc.html].
```
kou, line 109: Hmm. If we do this, it doesn't refer a file in the same version...

BurdetteLamar: Sorry, I meant to remove this change.  I've submitted https://github.com/ruby/csv/issues/284 for this link.

BurdetteLamar: This link is now removed in this PR.  The recipes are not available in ruby/ruby.

kou: OK.

PR #287, opened 2023-09-12
Add CSV::InvalidEncodingError

shibaaaa:
To handle encoding errors in CSV parsing with the appropriate error class

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -854,6 +854,15 @@ class CSV
     end
   end
 
+  # The error thrown when the parser encounters invalid encoding in CSV.
+  class InvalidEncodingError < MalformedCSVError
+    attr_reader :encoding
+    def initialize(encoding, line_number)
+      @encoding = encoding
+      super("Invalid byte sequence in #{encoding}", line_number)
+    end
+  end
+
   #
   # A FieldInfo Struct contains details about a field's position in the data
   # source it was read from.  CSV will pass this Struct to some blocks that make
--- a/test/csv/interface/test_read.rb
+++ b/test/csv/interface/test_read.rb
@@ -113,11 +113,11 @@ class TestCSVInterfaceRead < Test::Unit::TestCase
       file << "\u{1F600},\u{1F601}"
     end
     CSV.open(@input.path, encoding: "EUC-JP") do |csv|
-      error = assert_raise(CSV::MalformedCSVError) do
+      error = assert_raise(CSV::InvalidEncodingError) do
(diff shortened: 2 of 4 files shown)


lib/csv.rb
```diff
@@ -854,6 +854,13 @@ def initialize(message, line_number)
     end
   end
 
+  # The error thrown when the parser encounters invalid encoding in CSV.
+  class InvalidEncodingError < MalformedCSVError
+    def initialize(encoding, line_number)
```
kou, line 859: How about adding a reader for the target encoding?
```suggestion
    attr_reader :encoding
    def initialize(encoding, line_number)
      @encoding = encoding
```

kou (approved):

lib/csv.rb
kou: We can't get the target encoding by setting it to `@encoding`. (See the above suggestion.)

test/csv/interface/test_read.rb
```diff
@@ -113,7 +113,7 @@ def test_open_encoding_invalid
       file << "\u{1F600},\u{1F601}"
     end
     CSV.open(@input.path, encoding: "EUC-JP") do |csv|
-      error = assert_raise(CSV::MalformedCSVError) do
+      error = assert_raise(CSV::InvalidEncodingError) do
```
kou, line 116: Could you also change the below `assert_equal(...)` to the following?

```ruby
assert_equal([Encoding::EUC_JP, "Invalid byte sequence in EUC-JP in line 1."],
             [error.encoding, error.message])

File: test/csv/test_encodings.rb [hunk 3]
```diff
@@ -280,7 +280,7 @@ def test_row_separator_detection_with_invalid_encoding
   def test_invalid_encoding_row_error
     csv = CSV.new("valid,x\rinvalid,\xF8\r".force_encoding("UTF-8"),
                   encoding: "UTF-8", row_sep: "\r")
-    error = assert_raise(CSV::MalformedCSVError) do
+    error = assert_raise(CSV::InvalidEncodingError) do
```
kou, line 283: ditto.

lib/csv.rb
shibaaaa: oops!

kou:
Thanks!

kou:
Do you want a new version soon?

shibaaaa:
@kou 
> Do you want a new version soon?

No I don't, thanks.

PR #288, opened 2023-10-01
ci: test on maintained ruby versions

igas:


kou:
Why? We want to support Ruby 2.5.0 or later as much as possible.
https://github.com/ruby/csv/blob/master/csv.gemspec#L57

igas:
@kou because they are EOL, but I'm happy to just add 3.2. Should I restore 2.x?

olleolleolle:
👋 [Rdoc](https://github.com/ruby/rdoc/blob/master/.github/workflows/test.yml#L9) uses an Action from the https://github.com/ruby/actions repo. It's pretty neat. That way, small edits to CI configuration are often automatic.

igas:
Nice one @olleolleolle! Happy to switch to it, I guess we just need to agree if we support EOL ruby versions with security vulnerabilities.

kou:
> Should I restore 2.x?

Yes.

> Happy to switch to it

I don't object it.

kou:
I close this in favor of #289.

PR #299, opened 2024-03-28
Use www.rfc-editor.org for RFC text.

hsbt:
See details https://github.com/ruby/ruby/pull/10391

We should use `www.rfc-editor.org` for RFC references. I update docs in csv.


kou:
Wow, I didn't know the site.

PR #302, opened 2024-05-04
Allow CSV.open with StringIO argument

### LINKED ISSUE #300: Read StringIO with foreach
## Proposal

I'd like to propose being able to open a `StringIO` `s` with `CSV.foreach(s)`, just as it's possible to open a file with `CSV.foreach(path)`.

I can investigate and open a PR if this behavior is desired.

## Current behavior
The documentation already says it should be possible to pass an IO object to `foreach`, but if an instance of `StringIO` is passed, we get:
```
no implicit conversion of StringIO into String (TypeError)
```

MarcPer:
Fix #300

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1605,7 +1605,11 @@ class CSV
       options.delete_if {|k, _| /newline\z/.match?(k)}
 
       if filename_or_io.is_a?(StringIO)
-        f = StringIO.new(filename_or_io.string, **file_opts)
+        f = if RUBY_VERSION >= "2.7"
+          StringIO.new(filename_or_io.string, **file_opts)
+        else
+          StringIO.new(filename_or_io.string)
+        end
       else
         begin
           f = File.open(filename_or_io, mode, **file_opts)
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1605,7 +1605,7 @@ class CSV
       options.delete_if {|k, _| /newline\z/.match?(k)}
 
       if filename_or_io.is_a?(StringIO)
-        f = init_stringio!(filename_or_io.string, mode, **file_opts)
+        f = create_stringio(filename_or_io.string, mode, **file_opts)
       else
         begin
           f = File.open(filename_or_io, mode, **file_opts)
@@ -1924,13 +1924,13 @@ class CSV
     end
 
     if RUBY_VERSION < "2.7"
-      def init_stringio!(str, mode, opts)
-        opts.reject! { |k, _| k == :universal_newline || DEFAULT_OPTIONS.key?(k) }
+      def create_stringio(str, mode, opts)
+        opts.delete_if { |k, _| k == :universal_newline or DEFAULT_OPTIONS.key?(k) }
         raise ArgumentError, "Unsupported options parsing StringIO: #{opts.keys}" unless opts.empty?
-        StringIO.new(str)
+        StringIO.new(str, mode)
       end
     else
-      def init_stringio!(str, mode, opts)
+      def create_stringio(str, mode, opts)
         StringIO.new(str, mode, **opts)
       end
     end
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1508,10 +1508,8 @@ class CSV
 
     #
     # :call-seq:
-    #   open(file_path, mode = "rb", **options ) -> new_csv
-    #   open(io, mode = "rb", **options ) -> new_csv
-    #   open(file_path, mode = "rb", **options ) { |csv| ... } -> object
-    #   open(io, mode = "rb", **options ) { |csv| ... } -> object
+    #   open(path_or_io, mode = "rb", **options ) -> new_csv
+    #   open(path_or_io, mode = "rb", **options ) { |csv| ... } -> object
     #
     # possible options elements:
     #   keyword form:
@@ -1520,7 +1518,7 @@ class CSV
     #     :undef => :replace   # replace undefined conversion
     #     :replace => string   # replacement string ("?" or "\uFFFD" if not specified)
     #
-    # * Argument +path+, if given, must be the path to a file.
+    # * Argument +path_or_io+, must be a file path or an \IO stream.
     # :include: ../doc/csv/arguments/io.rdoc
     # * Argument +mode+, if given, must be a \File mode.
     #   See {Access Modes}[https://docs.ruby-lang.org/en/master/File.html#class-File-label-Access+Modes].
@@ -1544,6 +1542,10 @@ class CSV
     #   path = 't.csv'
     #   File.write(path, string)
     #
+    #   string_io = StringIO.new
+    #   string_io << "foo,0\nbar,1\nbaz,2\n"
+    #   string_io.rewind
+    #
     # ---
     #
     # With no block given, returns a new \CSV object.
@@ -1556,6 +1558,9 @@ class CSV
     #   csv = CSV.open(File.open(path))
     #   csv # => #<CSV io_type:File io_path:"t.csv" encoding:UTF-8 lineno:0 col_sep:"," row_sep:"\n" quote_char:"\"">
     #
+    # Create a \CSV object using a \StringIO:
+    #   csv = CSV.open(string_io)
+    #   csv # => #<CSV io_type:StringIO encoding:UTF-8 lineno:0 col_sep:"," row_sep:"\n" quote_char:"\"">
     # ---
     #
     # With a block given, calls the block with the created \CSV object;
@@ -1573,12 +1578,17 @@ class CSV
     # Output:
     #   #<CSV io_type:File io_path:"t.csv" encoding:UTF-8 lineno:0 col_sep:"," row_sep:"\n" quote_char:"\"">
     #
+    # Using a \StringIO:
+    #   csv = CSV.open(string_io) {|csv| p csv}
+    #   csv # => #<CSV io_type:StringIO encoding:UTF-8 lineno:0 col_sep:"," row_sep:"\n" quote_char:"\"">
+    # Output:
+    #   #<CSV io_type:StringIO encoding:UTF-8 lineno:0 col_sep:"," row_sep:"\n" quote_char:"\"">
     # ---
     #
     # Raises an exception if the argument is not a \String object or \IO object:
     #   # Raises TypeError (no implicit conversion of Symbol into String)
     #   CSV.open(:foo)
-    def open(filename, mode="r", **options)
+    def open(filename_or_io, mode="r", **options)
       # wrap a File opened with the remaining +args+ with no newline
       # decorator
       file_opts = {}
@@ -1592,14 +1602,19 @@ class CSV
       options.delete(:replace)
       options.delete_if {|k, _| /newline\z/.match?(k)}
 
-      begin
-        f = File.open(filename, mode, **file_opts)
-      rescue ArgumentError => e
-        raise unless /needs binmode/.match?(e.message) and mode == "r"
-        mode = "rb"
-        file_opts = {encoding: Encoding.default_external}.merge(file_opts)
-        retry
+      if filename_or_io.is_a?(StringIO)
+        f = filename_or_io
+      else
+        begin
+          f = File.open(filename_or_io, mode, **file_opts)
+        rescue ArgumentError => e
+          raise unless /needs binmode/.match?(e.message) and mode == "r"
+          mode = "rb"
+          file_opts = {encoding: Encoding.default_external}.merge(file_opts)
+          retry
+        end
       end
+
       begin
         csv = new(f, **options)
       rescue Exception
@@ -1612,7 +1627,7 @@ class CSV
         begin
           yield csv
         ensure
-          csv.close
+          csv.close unless f.is_a?(StringIO)
         end
       else
         csv
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1605,8 +1605,7 @@ class CSV
       options.delete_if {|k, _| /newline\z/.match?(k)}
 
       if filename_or_io.is_a?(StringIO)
-        filter_supported_stringio_open_opts!(file_opts)
-        f = StringIO.new(filename_or_io.string, **file_opts)
+        f = init_stringio!(filename_or_io.string, mode, **file_opts)
       else
         begin
           f = File.open(filename_or_io, mode, **file_opts)
@@ -1925,12 +1924,14 @@ class CSV
     end
 
     if RUBY_VERSION < "2.7"
-      def filter_supported_stringio_open_opts!(opts)
+      def init_stringio!(str, mode, opts)
         opts.reject! { |k, _| k == :universal_newline || DEFAULT_OPTIONS.key?(k) }
         raise ArgumentError, "Unsupported options parsing StringIO: #{opts.keys}" unless opts.empty?
+        StringIO.new(str)
       end
     else
-      def filter_supported_stringio_open_opts!(opts)
+      def init_stringio!(str, mode, opts)
+        StringIO.new(str, mode, **opts)
       end
     end
   end
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1605,11 +1605,8 @@ class CSV
       options.delete_if {|k, _| /newline\z/.match?(k)}
 
       if filename_or_io.is_a?(StringIO)
-        f = if RUBY_VERSION >= "2.7"
-          StringIO.new(filename_or_io.string, **file_opts)
-        else
-          StringIO.new(filename_or_io.string)
-        end
+        filter_supported_stringio_open_opts!(file_opts)
+        f = StringIO.new(filename_or_io.string, **file_opts)
       else
         begin
           f = File.open(filename_or_io, mode, **file_opts)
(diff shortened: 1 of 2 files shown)


lib/csv.rb
```diff
@@ -1590,14 +1600,19 @@ def open(filename, mode="r", **options)
...
 
-      begin
-        f = File.open(filename, mode, **file_opts)
-      rescue ArgumentError => e
-        raise unless /needs binmode/.match?(e.message) and mode == "r"
-        mode = "rb"
-        file_opts = {encoding: Encoding.default_external}.merge(file_opts)
-        retry
+      if filename_or_io.is_a?(StringIO)
+        f = filename_or_io
```
kou, line 1604: How about creating a new `StringIO` instead of reusing the given `StringIO` like we did for `IO`?

MarcPer: Good point, I missed that the `File.open` call to an already open file stream creates a new IO.

test/csv/interface/test_read.rb
```diff
@@ -48,6 +60,25 @@ def test_foreach_in_ractor
       ]
       assert_equal(rows, ractor.take)
     end
+
+    def test_foreach_stringio_in_ractor
```
kou, line 64:
```suggestion
    ractor
    def test_foreach_stringio_in_ractor
```

lib/csv.rb
```diff
@@ -1590,14 +1600,21 @@ def open(filename, mode="r", **options)
...
-        f = File.open(filename, mode, **file_opts)
-      rescue ArgumentError => e
-        raise unless /needs binmode/.match?(e.message) and mode == "r"
-        mode = "rb"
-        file_opts = {encoding: Encoding.default_external}.merge(file_opts)
-        retry
+      if filename_or_io.is_a?(StringIO)
+        pos = filename_or_io.pos
+        f = StringIO.new(filename_or_io.read)
+        filename_or_io.seek(pos)
```
kou, line 1606: Can we use `StringIO#string` here?

```suggestion
        f = StringIO.new(filename_or_io.string)
```

MarcPer: We'd then run into an issue similar to #98, which I confirmed with another test:

```ruby
 def test_foreach_stringio_with_bom
   s = StringIO.new
   s << "\ufeff"  # BOM
   s << @data
   s.rewind
   s.read(3)
   rows = CSV.foreach(s, col_sep: "\t", row_sep: "\r\n").to_a
   assert_equal(@rows, rows)
end
```

test/csv/interface/test_read.rb
MarcPer: I think this test will need to be removed, since `StringIO` cannot be copied or moved to the Ractor.

lib/csv.rb
MarcPer: I guess the most relevant counter-argument is, `CSV.new` also uses `read`, so using `read` here is also a matter of consistency.

kou: Good point. Could you also add a test for the BOM case?

How about using `StringIO#set_encoding_by_bom` for it?

```diff
        f = StringIO.new(filename_or_io.string)
        unless f.set_encoding_by_bom
          f.set_encoding(filename_or_io.external_encoding)
        end
```

`CSV.new` uses already opened IO. So it must not ignore the current position. So we must use `read`.
But `CSV.open` opens a new IO. So it must use all contents instead of contents after the current position.
Could you check the current behavior of `CSV.open` with sought real IO?

MarcPer: > But CSV.open opens a new IO. So it must use all contents instead of contents after the current position.
You're right, `CSV.open` reads the whole file content. I changed `open` again to use `string` instead of `read`.

> Could you also add a test for the BOM case?
Sure thing. I added it, and together with using the `**file_opts` in the new `StringIO`, it's handled correctly.

> How about using StringIO#set_encoding_by_bom for it?
I couldn't see any effect of doing it. It seems like including the `file_opts` is enough. Or is there a case I'm missing?

test/csv/interface/test_read.rb
```diff
@@ -32,6 +32,25 @@ def test_foreach
     assert_equal(@rows, rows)
   end
 
+  def test_foreach_stringio
+    s = StringIO.new
+    s << @data
```
kou, line 37: Can we simplify this?

```suggestion
    s = StringIO.new(@data)
```

test/csv/interface/test_read.rb
```diff
@@ -32,6 +32,25 @@ def test_foreach
     assert_equal(@rows, rows)
   end
 
+  def test_foreach_stringio
+    s = StringIO.new
```
kou, line 36: Could you use `string_io` instead of `s`? We don't need to abbreviate it here. 

```suggestion
    string_io = StringIO.new
```

test/csv/interface/test_read.rb
```diff
@@ -32,6 +32,25 @@ def test_foreach
...
+    CSV.foreach(s, col_sep: "\t", row_sep: "\r\n") do |row|
+      rows << row
+    end
+    assert_equal(@rows, rows)
+  end
+
+  def test_foreach_stringio_with_bom
+    s = StringIO.new
+    s << "\ufeff" # U+FEFF ZERO WIDTH NO-BREAK SPACE
+    s << @data
```
kou, line 49: Can we simplify this?

```suggestion
    s = StringIO.new("\ufeff#{@data}") # U+FEFF ZERO WIDTH NO-BREAK SPACE
```

lib/csv.rb
kou: > I couldn't see any effect of doing it.
 
 Ah, b706d91de8ec640d2918a3b4ff8828bcd6d0f09e is a trick for it.

MarcPer: Is there anything you still feel is missing, after I've added `file_opts` with the trick from https://github.com/ruby/csv/commit/b706d91de8ec640d2918a3b4ff8828bcd6d0f09e ?

test/csv/interface/test_read.rb
```diff
@@ -32,6 +32,22 @@ def test_foreach
     assert_equal(@rows, rows)
   end
 
+  def test_foreach_stringio
+    string_io = StringIO.new(@data)
+
+    rows = []
+    CSV.foreach(string_io, col_sep: "\t", row_sep: "\r\n") do |row|
+      rows << row
+    end
```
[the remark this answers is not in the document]
MarcPer: Sure thing! Done.

test/csv/interface/test_read.rb
```diff
@@ -32,6 +32,20 @@ def test_foreach
     assert_equal(@rows, rows)
   end
 
+  def test_foreach_stringio
+    string_io = StringIO.new(@data)
+
+    rows = []
```
kou, line 38:
```suggestion
```

kou:
Could you rebase on main? Our CI on main broken... and I've fixed it now.

MarcPer:
> Could you rebase on main? Our CI on master broken... and I've fixed it now.

Thank you, I've rebase it now.

kou:
Thanks. But sorry... The master CI was broken... I've fixed it. Could you rebase on master again?

MarcPer:
> Thanks. But sorry... The master CI was broken... I've fixed it. Could you rebase on master again?

No problem, I rebased it now.

MarcPer:
I'm assuming the test failures are related to https://bugs.ruby-lang.org/issues/20526, right?

kou:
Hmm. It should be fixed by https://github.com/ruby/csv/commit/47bf76fe7fb3f990802519a2f3af96fbcdc432b5 .
Could you rebase again?

MarcPer:
Rebased it, and removed a now unneeded check when closing the IO within `CSV#open`.

kou:
Ah, the failed test is an added test.

Does this work?

```diff
diff --git a/lib/csv.rb b/lib/csv.rb
index 7511ced..d2d1c57 100644
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1591,7 +1591,10 @@ class CSV
       # wrap a File opened with the remaining +args+ with no newline
       # decorator
       file_opts = {}
-      may_enable_bom_deletection_automatically(mode, options, file_opts)
+      may_enable_bom_deletection_automatically(filename_or_io,
+                                               mode,
+                                               options,
+                                               file_opts)
       file_opts.merge!(options)
       unless file_opts.key?(:newline)
         file_opts[:universal_newline] ||= false
@@ -1900,10 +1903,15 @@ class CSV
     private_constant :ON_WINDOWS
 
     private
-    def may_enable_bom_deletection_automatically(mode, options, file_opts)
-      # "bom|utf-8" may be buggy on Windows:
-      # https://bugs.ruby-lang.org/issues/20526
-      return if ON_WINDOWS
+    def may_enable_bom_deletection_automatically(filename_or_io,
+                                                 mode,
+                                                 options,
+                                                 file_opts)
+      unless filename_or_io.is_a?(StringIO)
+        # "bom|utf-8" may be buggy on Windows:
+        # https://bugs.ruby-lang.org/issues/20526
+        return if ON_WINDOWS
+      end
       return unless Encoding.default_external == Encoding::UTF_8
       return if options.key?(:encoding)
       return if options.key?(:external_encoding)
```

test/csv/interface/test_read.rb
```diff
@@ -32,6 +32,22 @@ def test_foreach
...
+      rows = CSV.foreach(string_io, col_sep: "\t", row_sep: "\r\n").to_a
+      assert_equal(@rows, rows)
+    end
+
+    def test_foreach_stringio_with_bom
+      string_io = StringIO.new("\ufeff#{@data}") # U+FEFF ZERO WIDTH NO-BREAK SPACE
+      rows = CSV.foreach(string_io, col_sep: "\t", row_sep: "\r\n").to_a
+      assert_equal(@rows, rows)
+    end
+  end
```
kou, line 49: Does `test_foreach_stringio` work with Ruby 2.7? If so, could you omit only `test_

```suggestion
  def test_foreach_stringio
    string_io = StringIO.new(@data)
    rows = CSV.foreach(string_io, col_sep: "\t", row_sep: "\r\n").to_a
    assert_equal(@rows, rows)
  end

  def test_foreach_stringio_with_bom
    if RUBY_VERSION < "2.7"
      # Support to StringIO was dropped for Ruby 2.6 and earlier without BOM support:
      # https://github.com/ruby/stringio/pull/47
      omit("StringIO's BOM support isn't available with Ruby < 2.7") 
    end
    string_io = StringIO.new("\ufeff#{@data}") # U+FEFF ZERO WIDTH NO-BREAK SPACE
    rows = CSV.foreach(string_io, col_sep: "\t", row_sep: "\r\n").to_a
    assert_equal(@rows, rows)
  end
```

MarcPer: It does if I also initialize `StringIO` without additional options. I just pushed another commit like that. It's unfortunate having to run an additional `RUBY_VERSION` check at runtime, but at least that's only when calling `open`.

lib/csv.rb
```diff
@@ -1592,14 +1604,23 @@ def open(filename, mode="r", **options)
...
-        raise unless /needs binmode/.match?(e.message) and mode == "r"
-        mode = "rb"
-        file_opts = {encoding: Encoding.default_external}.merge(file_opts)
-        retry
+      if filename_or_io.is_a?(StringIO)
+        f = if RUBY_VERSION >= "2.7"
+          StringIO.new(filename_or_io.string, **file_opts)
+        else
+          StringIO.new(filename_or_io.string)
+        end
```
kou, line 1612: Can we always use an empty `file_opts` for Ruby 2.7 or earlier instead?

```diff
diff --git a/lib/csv.rb b/lib/csv.rb
index 5f75894..6f94bd4 100644
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1605,11 +1605,7 @@ class CSV
       options.delete_if {|k, _| /newline\z/.match?(k)}
 
       if filename_or_io.is_a?(StringIO)
-        f = if RUBY_VERSION >= "2.7"
-          StringIO.new(filename_or_io.string, **file_opts)
-        else
-          StringIO.new(filename_or_io.string)
-        end
+        f = StringIO.new(filename_or_io.string, **file_opts)
       else
         begin
           f = File.open(filename_or_io, mode, **file_opts)
@@ -1911,7 +1907,11 @@ class CSV
                                                mode,
                                                options,
                                                file_opts)
-      unless filename_or_io.is_a?(StringIO)
+      if filename_or_io.is_a?(StringIO)
+        # Support to StringIO was dropped for Ruby 2.6 and earlier without BOM support:
+        # https://github.com/ruby/stringio/pull/47
+        return if RUBY_VERSION < "3.0"
+      else
         # "bom|utf-8" may be buggy on Windows:
         # https://bugs.ruby-lang.org/issues/20526
         return if ON_WINDOWS
```

MarcPer: This doesn't work because `CSV#open` still adds keys to `file_opts` here: https://github.com/ruby/csv/blob/bb93c289a529c77b5d5a8788dd8fdc03052216fc/lib/csv.rb#L1586-L1589

For Ruby < 2.7, if `file_opts` is not empty, we get `TypeError: no implicit conversion of Hash into String`.

I like that something fails if `options` is given to `CSV#open` when it's not supported, as opposed to simply ignoring it, the way I did it, but then the error should be ideally more clear than what the old `StringIO` returns. I can go that route, by rescuing the `TypeError`, and raising another exception with a clearer message. In any case, the `file_opts[:universal_newline] ||= false` would still have to be removed for `< 2.7`.

kou: > This doesn't work because `CSV#open` still adds keys to `file_opts` here:
> 
> https://github.com/ruby/csv/blob/bb93c289a529c77b5d5a8788dd8fdc03052216fc/lib/csv.rb#L1586-L1589

Oh, sorry.

> For Ruby < 2.7, if `file_opts` is not empty, we get `TypeError: no implicit conversion of Hash into String`.
> 
> I like that something fails if `options` is given to `CSV#open` when it's not supported, as opposed to simply ignoring it, the way I did it, but then the error should be ideally more clear than what the old `StringIO` returns. I can go that route, by rescuing the `TypeError`, and raising another exception with a clearer message.

I agree with you but can we use `raise ... unless file_opts.empty?` for it instead of rescuing the `TypeError`?

> In any case, the `file_opts[:universal_newline] ||= false` would still have to be removed for `< 2.7`.

I agree with you.

MarcPer: It turns out passing an empty `file_opts` doesn't work either. I've noticed this quirk from the interpreter in Ruby 2.6:

```ruby
StringIO.new("bla", **{}) # => #<StringIO:0x0000563108bb88b8>

opts = {}
StringIO.new("bla", **opts) # => TypeError (no implicit conversion of Hash into String)
```

To address that, I've changed the `StringIO` initialization in Ruby < 2.7, so that it:
- ignores options that are meant only for `CSV.new`
- raises an `ArgumentError` if anything else is passed as `options` to `CSV.open`
- initializes `StringIO` without keyword arguments.

lib/csv.rb
```diff
@@ -1886,16 +1903,37 @@ def table(path, **options)
...
+      end
       return unless Encoding.default_external == Encoding::UTF_8
       return if options.key?(:encoding)
       return if options.key?(:external_encoding)
       return if mode.include?(":")
       file_opts[:encoding] = "bom|utf-8"
     end
+
+    if RUBY_VERSION < "2.7"
+      def init_stringio!(str, mode, opts)
```
kou, line 1927: Could you remove `!` to indicate that this may raise an exception?
It's a Ruby on Rails convention not Ruby.

```suggestion
      def init_stringio(str, mode, opts)
```

BTW, how about renaming this to `create_stringio` or something?
Because this creates a new `StringIO` instead initializing existing `StringIO`.

lib/csv.rb
```diff
@@ -1886,16 +1903,37 @@ def table(path, **options)
...
+        opts.reject! { |k, _| k == :universal_newline || DEFAULT_OPTIONS.key?(k) }
```
kou, line 1928: Could you use this style to align with the existing code?

```suggestion
        opts.delete_if {|k, _| k == :universal_newline or DEFAULT_OPTIONS.key?(k)}
```

lib/csv.rb
kou: OK. Let's use the approach.

lib/csv.rb
MarcPer: Thank you for the feedback. I've followed all your suggestions.

lib/csv.rb
```diff
@@ -1886,16 +1903,37 @@ def table(path, **options)
...
       return unless Encoding.default_external == Encoding::UTF_8
       return if options.key?(:encoding)
       return if options.key?(:external_encoding)
       return if mode.include?(":")
       file_opts[:encoding] = "bom|utf-8"
     end
+
+    if RUBY_VERSION < "2.7"
+      def create_stringio(str, mode, opts)
+        opts.delete_if { |k, _| k == :universal_newline or DEFAULT_OPTIONS.key?(k) }
```
kou, line 1928:
```suggestion
        opts.delete_if {|k, _| k == :universal_newline or DEFAULT_OPTIONS.key?(k)}
```

kou:
Thanks.

PR #303, opened 2024-05-05
Move development dependencies to Gemfile

nobu:


csv.gemspec
```diff
@@ -57,8 +57,4 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = ">= 2.5.0"
 
   # spec.add_dependency "stringio", ">= 0.1.3"
```
kou, line 59:
```suggestion
```

Issue #301, opened 2024-05-04
feature request: Warn when reading BOM text with headers option


JunichiIto:
When BOM exists in CSV text, you cannot read first column by field name:

```ruby
require 'csv'

text = "\u{feff}id,name\n1,Alice"
csv = CSV.parse(text, headers: true)
csv[0]['id']   #=> nil
csv[0]['name'] #=> "Alice"
```

I understand there are some workarounds for this problem:

- Remove BOM from text before parse
- Include BOM to get value (eg. `csv[0]["\u{feff}id"]`)
- Specify encoding like `bom|utf-8` for `CSV.foreach`, `CSV.read`, `CSV.open` etc.

But the biggest problem is it's too hard to notice the existence of BOM. Actually, when I came across this issue, I had no idea why I failed to get the id value. I guess most people never wants to keep BOM in field name.

So I'd be happy if `CSV.parse` or some other reading methods would warn when input text contains BOM and `headers: true` is specified, like this:

```ruby
text = "\u{feff}id,name\n1,Alice"
csv = CSV.parse(text, headers: true)
#=> warning: first header name starts with U+FEFF (zero width no-break space, BOM)
```

```ruby
csv = CSV.read('with-bom.csv', headers: true)
#=> warning: first header name starts with U+FEFF (zero width no-break space, BOM)

# warning is not required when encoding like "bom|utf-8" is specified
csv = CSV.read('with-bom.csv', headers: true, encoding: 'bom|utf-8')
#=> (no warnings)
```

By the way, why do we put BOM so often? This is because Excel cannot open UTF-8 CSV correctly without BOM!
[Save CSV file in UTF\-8 with BOM\. Fix Korean text from being corrupted… \| by Hyunbin \| Medium](https://hyunbinseo.medium.com/save-csv-file-in-utf-8-with-bom-29abf608e86e)

Related issue: https://github.com/ruby/csv/issues/43

kou:
I can understand the motivation but I'm not sure that this is a proper approach.

How about enabling BOM detection in `CSV.open` by default instead?

JunichiIto:
> How about enabling BOM detection in `CSV.open` by default instead?

Do you mean BOM is removed automatically without specifying encoding?

```ruby
# like this
csv = CSV.open('with-bom.csv', headers: true)
csv[0]['id'] #=> 1
```

It would be much better than warning!

kou:
Yes.

How about this?

```diff
[...]

PR #305, opened 2024-05-17
doc CSV.parse: Add BOM handling

JunichiIto:
Requested in https://github.com/ruby/csv/issues/301#issuecomment-2114358930

Screenshot 2024-05-17 at 10 48 23


kou:
Thanks.

PR #307, opened 2024-06-13
Fix typo in recipe "Parse from File with Headers"

herwinw:
The text mentioned an instance method, the code itself uses a class method.

### DIFF
--- a/doc/csv/recipes/parsing.rdoc
+++ b/doc/csv/recipes/parsing.rdoc
@@ -110,7 +110,7 @@ You can parse \CSV data from a \File, with or without headers.
 
 ===== Recipe: Parse from \File with Headers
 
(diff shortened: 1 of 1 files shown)


doc/csv/recipes/parsing.rdoc
```diff
@@ -110,7 +110,7 @@ You can parse \CSV data from a \File, with or without headers.
 
 ===== Recipe: Parse from \File with Headers
 
-Use instance method CSV#read with option +headers+ to read a file all at once:
+Use class method CSV#read with option +headers+ to read a file all at once:
```
kou, line 113:
```suggestion
Use class method CSV.read with option +headers+ to read a file all at once:
```

kou:
Good catch!

PR #313, opened 2024-08-18
Add Time converter to CSV::Converters

bdewater:
Ruby recommends working with Time objects, unless you have a good reason to use DateTime: https://ruby-doc.org/stdlib-2.5.0/libdoc/date/rdoc/DateTime.html#class-DateTime-label-When+should+you+use+DateTime+and+when+should+you+use+Time-3F

This commit adds the missing converter for the common modern use case.

### DIFF
--- a/doc/csv/recipes/parsing.rdoc
+++ b/doc/csv/recipes/parsing.rdoc
@@ -45,6 +45,7 @@ All code snippets on this page assume that the following has been executed:
     - {Recipe: Convert Fields to Numerics}[#label-Recipe-3A+Convert+Fields+to+Numerics]
     - {Recipe: Convert Fields to Dates}[#label-Recipe-3A+Convert+Fields+to+Dates]
     - {Recipe: Convert Fields to DateTimes}[#label-Recipe-3A+Convert+Fields+to+DateTimes]
+    - {Recipe: Convert Fields to Times}[#label-Recipe-3A+Convert+Fields+to+Times]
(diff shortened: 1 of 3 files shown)


test/csv/test_data_converters.rb
```diff
@@ -187,4 +187,146 @@ def test_builtin_date_time_converter_rfc3339_tab_utc
     assert_equal(datetime,
                  CSV::Converters[:date_time][rfc3339_string])
   end
+
+  def test_builtin_time_converter
+    # does convert
+    assert_instance_of(Time,
+                       CSV::Converters[:time][@win_safe_time_str] )
```
kou, line 194:
```suggestion
                       CSV::Converters[:time][@win_safe_time_str])
```

doc/csv/recipes/parsing.rdoc
```diff
@@ -381,6 +382,13 @@ Convert fields to \DateTime objects using built-in converter +:date_time+:
   parsed = CSV.parse(source, headers: true, converters: :date_time)
   parsed.map {|row| row['DateTime'].class} # => [DateTime, DateTime, DateTime]
 
+===== Recipe: Convert Fields to Times
+
+Convert fields to \DateTime objects using built-in converter +:time+:
```
kou, line 387:
```suggestion
Convert fields to \Time objects using built-in converter +:time+:
```

bdewater:
Addressed PR feedback and the CI failures on Ruby 2.5 and 2.6.

kou:
Thanks.

Issue #272, opened 2023-01-27
 Suggestion to add `sep` option


kojix2:
Hi! CSV developers!

Currently, the `col_sep` option is used in Ruby's CSV library, 

```ruby
CSV.open("nice.tsv", col_sep: "\t")
```

but it requires 8 keystrokes. `sep` requires only 3 keystrokes.

The reason `col_sep` exists is because the `row_sep` option exists. 
However, `row_sep` is used much less frequently than `col_sep`.

Using `sep` as an alias for `col_sep` would make using the CSV library a little easier and more enjoyable.

Many R and Python libraries use `sep` in the sense of `col_sep`. For example, pandas can read a tsv file as follows

```python
import pandas as pd

df = pd.read_csv(file_path, sep="\t")
```

In pandas, `delimiter` is an alias for `sep`. It is not uncommon for `sep` options to have aliases in this way.

It took some courage to make such a proposal to a library with such a long history as CSV. 

Thanks for reading.

kou:
We may want to add `TSV` class that uses `\t` as the default column separator.

kou:
We can use `CSV::TSV.open("nice.tsv")`.

PR #317, opened 2024-10-28
Add :sep option as alias for :col_sep and TSV class for tab-separated values

jsxs0:
This PR adds two features to improve the usability of the CSV library:

1. Adds :sep as an alias for :col_sep
   - Provides a shorter option name (3 vs. 8 keystrokes)
   - Aligns with similar libraries like pandas (pd.read_csv(sep="\t"))
   - Maintains backward compatibility 
   - Raises clear error if both options specified

2. Adds TSV class to handle tab-separated files
   - Inherits from CSV class
   - Uses tab as default separator
   - Full interface parity with CSV
   - All class methods supported (parse, read, generate, etc.)
   - Allows override of default separator if needed

---
Example usage:
```ruby
# Using new :sep option
CSV.read("data.tsv", sep: "\t")

# Using new TSV class with default tab separator
TSV.read("data.tsv")
TSV.parse("a\tb\tc")
```

---

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -2033,6 +2033,7 @@ class CSV
   #
   def initialize(data,
                  col_sep: ",",
+                 sep: nil,
                  row_sep: :auto,
                  quote_char: '"',
                  field_size_limit: nil,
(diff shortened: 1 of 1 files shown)


kou:
Could you use separated PRs for multiple features?

jsxs0:
Continued at https://github.com/ruby/csv/pull/318, https://github.com/ruby/csv/pull/319

PR #318, opened 2024-10-28
Add :sep as an alias for :col_sep

jsxs0:
This PR adds :sep as a shorter alias for :col_sep to improve usability:

- Adds :sep option that acts as alias for :col_sep
- Reduces keystrokes (3 vs 8) when specifying separator
- Aligns with other data libraries like pandas (pd.read_csv(sep="\t"))
- Maintains backward compatibility
- Raises clear error if both options specified

---
Example usage:
```ruby
# Old way
CSV.read("data.tsv", col_sep: "\t")

# New way
CSV.read("data.tsv", sep: "\t")
```

---

The change is motivated by:

1. Making the interface more ergonomic with shorter option name
2. Better alignment with common practices in other data libraries

---
Changes:

1. Adds sep parameter to CSV#initialize
2. Adds error check for both options
3. Uses sep value if provided, otherwise falls back to col_sep

---

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -2033,6 +2033,7 @@ class CSV
   #
   def initialize(data,
                  col_sep: ",",
+                 sep: nil,
                  row_sep: :auto,
                  quote_char: '"',
                  field_size_limit: nil,
@@ -2058,6 +2059,11 @@ class CSV
                  write_nil_value: nil,
                  write_empty_value: "")
     raise ArgumentError.new("Cannot parse nil as CSV") if data.nil?
+
+    if sep && col_sep != ","
+      raise ArgumentError, "Cannot specify both :sep and :col_sep"
+    end
+    actual_sep = sep || col_sep
 
     if data.is_a?(String)
       if encoding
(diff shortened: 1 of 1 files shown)


kou:
I'm negative this.
If we have a special API for TSV something like #319, we don't need to specify `col_sep:` explicitly in most use cases?

kojix2:
I agree with @jsxs0.
I still like the short appearance of `sep` and do not like the naming of `col_sep`.
However, I am not going to argue strongly about this. I am not a regular user of the CSV library.

kou:
OK.
Let's close this.

kojix2:
I would like you to explain why this PR was closed.
I think there are at least 100 people on the planet who would like this change using Ruby's CSV library. Or maybe more. These people may search for pull requests. I don't think the idea behind this decision has been sufficiently explained.

kou:
> Consistency with row_sep

This is the reason.

kou:
> there are many file formats that use “|” as a delimiter.

It will be rare than TSV. Because there is no Pipe-Separated-Values name or something.

PR #319, opened 2024-10-28
Add TSV class for tab-separated values

jsxs0:
This PR adds a new TSV class to provide first-class support for tab-separated files:

1. Inherits from CSV class
2. Uses tab as default separator
3. Full interface parity with CSV
4. All class methods supported (parse, read, generate etc.)
5. Allows override of default separator if needed

---
Example usage:

```
# Use TSV class with default tab separator
TSV.read("data.tsv")
TSV.parse("a\tb\tc")

# Override separator if needed
TSV.read("data.csv", col_sep: ",")
```

---
The change is motivated by:

1. Providing better support for TSV files
2. Making tab-separated file handling more convenient

---

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -2129,6 +2129,92 @@ class CSV
     writer if @writer_options[:write_headers]
   end
 
+  class TSV < CSV
+    def initialize(data, **options)
+      options = options.dup
+      if !options.key?(:col_sep) && !options.key?(:sep)
+        options[:col_sep] = "\t"
+      end
+      super(data, **options)
+    end
+
+    class << self
+      def foreach(path, mode="r", **options, &block)
+        options = options.dup
+        if !options.key?(:col_sep) && !options.key?(:sep)
+          options[:col_sep] = "\t"
+        end
+        CSV.foreach(path, mode, **options, &block)
+      end
+
+      def read(path, **options)
+        options = options.dup
+        if !options.key?(:col_sep) && !options.key?(:sep)
+          options[:col_sep] = "\t"
+        end
+        CSV.read(path, **options)
+      end
+
+      def parse(str, **options)
+        options = options.dup
+        if !options.key?(:col_sep) && !options.key?(:sep)
+          options[:col_sep] = "\t"
+        end
+        CSV.parse(str, **options)
+      end
+
+      def generate(str = nil, **options)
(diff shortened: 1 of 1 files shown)


kou:
Hmm. Do we need to override all class methods? Can we just override `TSV#initialize`?

```diff
diff --git a/lib/csv.rb b/lib/csv.rb
index f6eb32a..8c71b4d 100644
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -2979,6 +2979,12 @@ class CSV
   end
 end
 
+class TSV < CSV
+  def initialize(data, **options)
+    super(data, **({col_sep: "\t"}.merge(options)))
+  end
+end
+
 # Passes +args+ to CSV::instance.
 #
 #   CSV("CSV,data").read
```

BTW, is `CSV::TSV` intentional? (`::TSV` may be easy to use but it may break other library's API that defines `::TSV`...)

jsxs0:
@kou Thanks for the feedback! Updated and pushed a new commit.

kou:
Could you add some tests for this change?
Could you update the PR description for the latest changes?

jsxs0:
> Could you add some tests for this change? Could you update the PR description for the latest changes?

@kou Thanks, added tests and updated the description.

run-test.rb
kou, line 1: Why do we need to change this file?
I think that we don't need to change this file but if you think that we need to change this file, could you open a separated PR?

test/csv/test_tsv.rb
```diff
@@ -0,0 +1,32 @@
+require_relative "helper"
+
+class TestTSV < Test::Unit::TestCase
+  def test_default_separator
+    tsv = TSV.new(String.new)
```
kou, line 5:
```suggestion
    tsv = CSV::TSV.new(String.new)
```

kou:
Did you run added tests on your local machine? Or did you enable GitHub Actions on your fork?
 
We need to add `CSV::` to all `TSV` references.

jsxs0:
> Did you run added tests on your local machine? Or did you enable GitHub Actions on your fork?
> 
> We need to add `CSV::` to all `TSV` references.

Thanks. The fix was a straightforward namespace correction that aligns the test code with the actual implementation in `lib/csv.rb`.

kou:
Thanks.

PR #320, opened 2024-11-04
CLI for filter

BurdetteLamar:
Adds a command-line interface for CSV::filter;  this supports using CSV without having to write Ruby code.

In its present form, this is a true filter:  reads from stdin, writes to stdout.  More could be done: input could be from ARGF or an IO stream;  output could be to IO stream.


BurdetteLamar:
I'm not sure what to do about these failures.  I chose to `require 'minitest/autorun'` because it has the assertion method `capture_subprocess_io`, which I've used to evaluate the output.

Suggestions welcome.

kou (commented):
This is an interested suggestion. But `filter` is too generic for command line name to handle only CSV.

How about adding `csv-` prefix or something?

test/csv/filter/test_filter.rb
```diff
@@ -0,0 +1,714 @@
+# -*- coding: utf-8 -*-
+# frozen_string_literal: false
+
+require 'minitest/autorun'
```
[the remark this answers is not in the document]
BurdetteLamar: Adapted.

BurdetteLamar:
> This is an interested suggestion. But `filter` is too generic for command line name to handle only CSV.
> 
> How about adding `csv-` prefix or something?

Changed to `csv-filter`.

BurdetteLamar:
Should have added earlier:  I've needed substantial debugging code to get this work done, and have left it in (just in case).

bin/csv-filter
```diff
@@ -0,0 +1,244 @@
...
+end
+
+options = {}
+
+OptionParser.new do |p|
+
+  p.program_name = File.basename $0
+  p.version = CSV::VERSION
+  p.release = nil
+  p.summary_indent = ' ' * 4
```
kou, line 26: We can omit this.

bin/csv-filter
```diff
@@ -0,0 +1,244 @@
...
+  p.banner = <<-EOF
```
kou, line 27: Could you use `BANNER` not `EOF` for here document mark?
`EOF` (End Of File?) is difficult to understand.

bin/csv-filter
```diff
@@ -0,0 +1,244 @@
...
+Usage: #{p.program_name} [options]
+
+  Reads and parses the CSV text content of $stdin' per the given input options.
```
kou, line 30: We don't need to use `$stdin` for the standard input because this program's users don't need to know about Ruby.

```suggestion
  Reads and parses the CSV text content of the standard input per the given input options.
```

bin/csv-filter
```diff
@@ -0,0 +1,244 @@
...
+  From that content, generates CSV text per the given output options and writes
+  that text to $stdout.
```
kou, line 32:
```suggestion
  that text to the standard output.
```

bin/csv-filter
```diff
@@ -0,0 +1,244 @@
...
+
+  Three options are similar for input and output:
+
+  - Row separator:
+  
+    - The input row separator is determined:
+  
+      - By option --input_row_sep, if given.
```
kou, line 40: Could you use `-` not `_` for word separator in option name: `--input-row-sep`

bin/csv-filter
```diff
@@ -0,0 +1,244 @@
...
+      - By option --output_row_sep if given.
+      - Otherwise, by option --row_sep if given.
+      - Otherwise, as the newline character ("\\n").
+  
+  - Column separator:
+  
+    - The input column separator is determined:
+  
+      - By option --input_col_sep, if given.
+      - Otherwise, by option --col_sep``, if given.
```
kou, line 55:
```suggestion
      - Otherwise, by option --col_sep, if given.
```

bin/csv-filter
```diff
@@ -0,0 +1,244 @@
...
+      - Otherwise, as the comma character (',').
```
kou, line 56: The above explanation uses `"` not `'` for quotation: `"\\n"`

```suggestion
      - Otherwise, as the comma character (",").
```

bin/csv-filter
```diff
@@ -0,0 +1,244 @@
...
+    - The output quote character is determined:
+  
+      - By option --output_quote_char, if given.
+      - Otherwise, by option --quote_char, if given.
+      - Otherwise, as the double-quote character ('"').
+EOF
+
+  p.separator(nil)
+  p.separator(
+    'Input Options: see https://docs.ruby-lang.org/en/3.3/CSV.html#class-CSV-label-Options+for+Parsing.'
```
kou, line 81: * We don't want to embed `3.3` in URL.
* Could you remove the last `.` to avoid that `.` is included in used URL?

bin/csv-filter
```diff
@@ -0,0 +1,244 @@
...
+  )
+
+  p.separator(nil)
+  p.on('--converters STR',
```
kou, line 85: * Could you use `--name=LABEL` style not `--name LABEL`?
* Could you use better name than `STR`?

```suggestion
  p.on('--converters=CONVERTERS',
```

bin/csv-filter
```diff
@@ -0,0 +1,244 @@
+#!/usr/bin/env ruby
+
+require 'optparse'
+require 'csv'
+
+# Returns an array of converter names.
+# \String +s+ contains a comma-separated list of converter names;
+# each optionally begins with a colon.
```
kou, line 8: Why do we need this feature?

bin/csv-filter
```diff
@@ -0,0 +1,244 @@
...
+def parse_converters(s)
+  converters = []
+  s.split(',').each do |name|
+    name.sub(/^:/, '')
+    sym = name.to_sym
+    converters.push(sym)
+  end
+  converters
+end
```
kou, line 17:
```suggestion
def parse_converters(raw_converters)
  raw_converters.split(",").collect(&:to_sym)
end
```

bin/csv-filter
```diff
@@ -0,0 +1,244 @@
...
+       ) do |value|
+    options[:converters] = parse_converters(value)
+  end
+  p.on('--empty_value STR',
+       'Value to be output for empty input field.') do |value|
+    options[:empty_value] = value
+  end
+  p.on('--field_size_limit N',
+       'Maximum field size (characters)') do |value|
+    options[:field_size_limit] = value.to_i
```
kou, line 97:
```suggestion
  p.on('--field_size_limit N', Integer,
       'Maximum field size (characters)') do |value|
    options[:field_size_limit] = value
```

bin/csv-filter
```diff
@@ -0,0 +1,244 @@
...
+  end
+  p.on('--headers STR',
+       'Treat the first input row as headers;',
+       'STR is true or comma-separated headers.'
+       ) do |value|
+    options[:headers] = value == 'true' ? true : parse_converters(value)
```
kou, line 103: It's strange that we use `parse_converters` for headers.
We need better name than `parse_converters`.

bin/csv-filter
```diff
@@ -0,0 +1,244 @@
...
+  end
+  p.on('--header_converters STR',
+       'Input header converters;',
+       'STR is comma-separated converter names (leading colon optional).'
+  ) do |value|
+    options[:header_converters] = parse_converters(value)
+  end
+  p.on('--liberal_parsing',
+       'Attempt to parse non-conformant input.') do
+    options[:liberal_parsing] = :no_argument
```
kou, line 113: `:no_argument` is strange. `:liberal_parsing` must be `true` or `false`: `p.on('--[no-]liberal-parsing', ...)`

kou (commented):
I didn't review all changes. Because there are many parts I need to add comments... Can we start from a smaller script instead of this?

BurdetteLamar:
Thanks, @kou, for your interest, and for this helpful review.  I'm at RubyConf in Chicago until the weekend, but will work on this when I get home.

When we're done with this review cycle, I'll be interested in adding to this work.  The filter will not be very useful without a config file (maybe similar to `~/.gitconfig`?) that will avoid the need for including lots of options on the command line.

BurdetteLamar:
I have commented out 13 of the options, leaving 16 in.  Each of those left in is either because it's mentioned in the help or because it has comments here.

bin/csv-filter
```diff
@@ -0,0 +1,244 @@
...
+    name.sub(/^:/, '')
+    sym = name.to_sym
+    converters.push(sym)
+  end
+  converters
+end
+
+options = {}
+
+OptionParser.new do |p|
```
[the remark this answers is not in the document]
BurdetteLamar: Done.

bin/csv-filter
```diff
@@ -0,0 +1,244 @@
...
+
+  p.program_name = File.basename $0
```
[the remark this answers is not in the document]
BurdetteLamar: Done.

bin/csv-filter
```diff
@@ -0,0 +1,244 @@
...
+  p.version = CSV::VERSION
+  p.release = nil
```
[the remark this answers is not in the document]
BurdetteLamar: Done.

bin/csv-filter
BurdetteLamar: Done.

bin/csv-filter
BurdetteLamar: Done.

bin/csv-filter
BurdetteLamar: Done.

bin/csv-filter
BurdetteLamar: Done.

bin/csv-filter
BurdetteLamar: Done (for those not commented out -- will get the rest as we go).

bin/csv-filter
BurdetteLamar: Fixed.

bin/csv-filter
BurdetteLamar: Changed (but I usually use double-quotes when interpolation is needed, single-quotes when not).

bin/csv-filter
BurdetteLamar: Dot removed.  I could not find a suitable replacement URL.  Suggestion?

bin/csv-filter
BurdetteLamar: Changed to equal-style for those options not commented out;  changed STR to something else.

bin/csv-filter
BurdetteLamar: Changed.

bin/csv-filter
BurdetteLamar: Replaced and renamed.

bin/csv-filter
BurdetteLamar: Fixed.

bin/csv-filter
BurdetteLamar: Renamed.

bin/csv-filter
BurdetteLamar: Symbol `:no_argument` is only internal, and tells the testing how to behave.  The CSV user does not see it.

bin/csv-filter
kou: We use double-quotes for all string literals. Could you follow our style in this repository?

kou:
Could you remove them from this PR instead of commenting them out?

FYI: You can copy the current branch content:

```bash
git clone git@github.com:BurdetteLamar/csv.git
cd csv
git switch filter
git switch -c filter-keep
git push origin filter-keep
```

BTW, can remove more options from this PR to make this script smaller?

bin/csv-filter
kou: Can we use `RUBY_VERSION.split(".", 3)[0, 2].join(".")` instead of `3.3`?

BTW, we should provide versioned documentation URL in https://ruby.github.io/csv/ something like:
* https://ruby.github.io/csv/master/ - this has the documents for the master branch
* https://ruby.github.io/csv/latest/ - this has the same content as https://ruby.github.io/csv/3.3.0/
* https://ruby.github.io/csv/3.3.0/
* https://ruby.github.io/csv/3.2.9/
* ...

BurdetteLamar:
> smaller

I'm thinking to close this PR and start fresh, with just `--help` and `--version`.  Then we can have a new PR for each option as we go, adjusting the help text and version accordingly.  Does that work for you?

kou:
Yes. It's better than the current PR.

BurdetteLamar:
Closing this;  will begin again with a new PR.

PR #322, opened 2024-11-27
Drop `:nodoc:` options from core classes that contains methods with docs

soutaro:
I found some methods are not included in RDoc output, and the `:nodoc:` option should be the reason.

It seems like there is no reason to opt-out the methods. https://github.com/ruby/csv/commit/7b79bec2779c53245cabe396ef36a13e8ab50338 added the option, but still not clear why we needed it.

### DIFF
--- a/lib/csv/core_ext/array.rb
+++ b/lib/csv/core_ext/array.rb
@@ -1,4 +1,4 @@
-class Array # :nodoc:
+class Array
   # Equivalent to CSV::generate_line(self, options)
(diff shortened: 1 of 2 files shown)


soutaro:
Oops I should have read this ticket. But still believe having the docs is better. https://bugs.ruby-lang.org/issues/6880

kou:
Thanks.

I don't have a strong opinion for this.

PR #339, opened 2025-03-23
csv-filter: Tweaks for help

BurdetteLamar:
In help schemas, change `SEPARATOR` to `SEP` (for readability).


kou:
-1

In general, no abbreviation is better for users.

PR #340, opened 2025-03-26
csv-filter: add --quote-char

BurdetteLamar:


test/csv/test_csv_filter.rb
```diff
@@ -137,4 +137,19 @@ def test_option_col_sep
...
                  run_csv_filter(csv, "--col-sep=:", "--output-col-sep=."))
   end
+
+  def test_option_quote_char
+    quote_char = "'"
+    csv = CSV.generate(quote_char: quote_char) do |csv|
+      csv << ['foo', 0]
+      csv << ["'bar'", 1]
+      csv << ['"baz"', 2]
+    end
```
kou, line 147:
```suggestion
    csv = "foo,0\n'''bar''',1\n\"baz\",2\n"
```

test/csv/test_csv_filter.rb
```diff
@@ -137,4 +137,19 @@ def test_option_col_sep
...
+    assert_equal(["foo,0\n'''bar''',1\n\"baz\",2\n", ""],
+                 run_csv_filter(csv, "--quote-char=:"))
```
kou, line 149: Is `--quote-char=:` a good test?
It does nothing, right?

test/csv/test_csv_filter.rb
```diff
@@ -137,4 +137,19 @@ def test_option_col_sep
...
+    assert_equal(["foo,0\n'''bar''',1\n\"\"\"baz\"\"\",2\n", ""],
+                 run_csv_filter(csv, "--input-quote-char=:"))
```
kou, line 151: Should we specify `--quote-char=...` too?

test/csv/test_csv_filter.rb
```diff
@@ -137,4 +137,19 @@ def test_option_col_sep
...
+    assert_equal(["foo,0\n'''bar''',1\nbaz,2\n", ""],
+                 run_csv_filter(csv, "--output-quote-char=:"))
```
kou, line 153: ditto.

test/csv/test_csv_filter.rb
BurdetteLamar: I think we have the same situation for `--col-sep=:` and `--row-sep=:`, don't we?  They all at least test that the option is accepted?

test/csv/test_csv_filter.rb
BurdetteLamar: I don't understand.  Doesn't the default value serve us well enough?

test/csv/test_csv_filter.rb
BurdetteLamar: Ditto.

test/csv/test_csv_filter.rb
kou: `--row-sep` tests use `--row-sep` and `--{input,output}-row-sep`.
I think that we want to test that `--quote-char` configuration is overridden by `--{input,output}-quote-char` here. Is it incorrect?

If we want to test only `--{input,output}-quote-char`, we should do it in the test for `--{input,output}-quote-char` not the test for `--quote-char`.

test/csv/test_csv_filter.rb
kou: Really?

They use `:` in the input CSV:

https://github.com/ruby/csv/blob/7eea19a639d6c03829b337e5c916804d8e1c10cc/test/csv/test_csv_filter.rb#L132-L134

https://github.com/ruby/csv/blob/7eea19a639d6c03829b337e5c916804d8e1c10cc/test/csv/test_csv_filter.rb#L122-L124

BurdetteLamar:
Maybe you should make these tests the way you want them?

kou:
Hmm.

It may be better that you create "csv-filter" gem and maintain it by yourself instead of maintaining `csv-filter` in "csv" gem.

BurdetteLamar:
> Hmm.
> 
> It may be better that you create "csv-filter" gem and maintain it by yourself instead of maintaining `csv-filter` in "csv" gem.

Okay with me;  I won't do more here.  I thought you might find this useful.

kou:
OK. I'll remove `bin/csv-filter` for now because we don't have enough maintenance resource for it.

Someone who wants to maintain can restart from https://github.com/ruby/csv/blob/7eea19a639d6c03829b337e5c916804d8e1c10cc/bin/csv-filter .

kou:
#341

PR #343, opened 2025-04-30
Replace link to ruby-doc.org with docs.ruby-lang.org

p8:
Link to official docs (ruby-doc.org is not official).

### DIFF
--- a/lib/csv/parser.rb
+++ b/lib/csv/parser.rb
@@ -47,7 +47,7 @@ class CSV
     # Uses StringScanner (the official strscan gem). Strscan provides lexical
     # scanning operations on a String. We inherit its object and take advantage
     # on the methods. For more information, please visit:
-    # https://ruby-doc.org/stdlib-2.6.1/libdoc/strscan/rdoc/StringScanner.html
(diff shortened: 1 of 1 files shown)


lib/csv/parser.rb
```diff
@@ -47,7 +47,7 @@ class UnexpectedError < StandardError
     # Uses StringScanner (the official strscan gem). Strscan provides lexical
     # scanning operations on a String. We inherit its object and take advantage
     # on the methods. For more information, please visit:
-    # https://ruby-doc.org/stdlib-2.6.1/libdoc/strscan/rdoc/StringScanner.html
+    # https://docs.ruby-lang.org/en/3.4/StringScanner.html
```
kou, line 50: How about using "master" here? We don't want to maintain the latest version...

```suggestion
    # https://docs.ruby-lang.org/en/master/StringScanner.html
```

kou: FYI: This is an internal class. (This is not for users.) So we can use "master" here.

p8:
Thanks @kou, I've made the changes.

kou:
Thanks.

PR #346, opened 2025-06-05
Fix: Allow \r in unquoted fields when row separator doesn't contain \r

### LINKED ISSUE #60: Accept \r without quote if row separator doesn't include \r
This may work:

```diff
@@ -360,12 +320,14 @@ class CSV
       if @liberal_parsing
         @unquoted_value = Regexp.new("[^".encode(@encoding) +
                                      escaped_column_separator +
-                                     "\r\n]+".encode(@encoding))
+                                     escaped_row_separator +
+                                     "]+".encode(@encoding))
       else
         @unquoted_value = Regexp.new("[^".encode(@encoding) +
                                      escaped_quote_character +
                                      escaped_column_separator +
-                                     "\r\n]+".encode(@encoding))
+                                     escaped_row_separator +
+                                     "]+".encode(@encoding))
       end
       @cr_or_lf = Regexp.new("[\r\n]".encode(@encoding))
       @not_line_end = Regexp.new("[^\r\n]+".encode(@encoding))
```

jsxs0:
Fixes #60

This has been bugging me for a while - the CSV parser was rejecting `\r` characters in unquoted fields even when the row separator was something completely different like `\n` or a custom separator.

For example, this would fail unnecessarily:
```ruby
CSV.parse("field1,field\rwith\rcr,field3\n", row_sep: "\n")
```

The problem was in `prepare_unquoted` where we were hardcoding `"\r\n"` instead of checking what the actual row separator was.

**What changed:**
- Now we only exclude characters that are actually part of the row separator
- If your row separator is `\n`, then `\r` is allowed in unquoted fields  
- If your row separator is `\r\n`, then both `\r` and `\n` are still properly excluded
- Quoted fields work exactly the same as before

**Testing:**
- Updated the tests that were expecting the old behavior
- Added comprehensive tests for different row separator scenarios
- All existing tests still pass

This makes the parser more flexible while keeping it safe for the cases where `\r` should actually be restricted.


lib/csv/parser.rb
```diff
@@ -675,7 +675,10 @@ def prepare_quoted
     def prepare_unquoted
       return if @quote_character.nil?
 
-      no_unquoted_values = "\r\n".encode(@encoding)
+      # Only exclude characters that are actually part of the row separator
+      # instead of hardcoding "\r\n"
+      row_separator_chars = @row_separator.chars.map { |c| Regexp.escape(c) }.join
```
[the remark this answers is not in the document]
kou: Why do we need `chars.map.join`?
Why can't we use `Regexp.escape(@row_separator)`?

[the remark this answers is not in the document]
jsxs0: Fixed! Using `Regexp.escape(@row_separator)?` now. Thanks!

lib/csv/parser.rb
```diff
@@ -675,7 +675,9 @@ def prepare_quoted
     def prepare_unquoted
       return if @quote_character.nil?
 
-      no_unquoted_values = "\r\n".encode(@encoding)
+              # Only exclude characters that are actually part of the row separator
+        # instead of hardcoding "\r\n"
+        no_unquoted_values = Regexp.escape(@row_separator).encode(@encoding)
```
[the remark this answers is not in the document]
jsxs0: Fixed! Thanks!

test/csv/parse/test_invalid.rb
```diff
@@ -1,17 +1,9 @@
...
 class TestCSVParseInvalid < Test::Unit::TestCase
-  def test_no_column_mixed_new_lines
-    error = assert_raise(CSV::MalformedCSVError) do
-      CSV.parse("\n" +
-                "\r")
-    end
-    assert_equal("New line must be <\"\\n\"> not <\"\\r\"> in line 2.",
-                 error.message)
-  end
+
```
kou, line 7:
```suggestion
```

test/csv/parse/test_general.rb
```diff
@@ -138,28 +138,71 @@ def test_non_regex_edge_cases
...
-  def test_malformed_csv_cr_middle_line
-    csv = <<-CSV
-line,1,abc
-line,2,"def\nghi"
+  def test_unquoted_cr_rejected_when_included_in_row_separator
+    data = "field1,field\r2,field3\r\nrow2,data,here\r\n"
+    assert_raise(CSV::MalformedCSVError) do
+      CSV.parse(data, row_sep: "\r\n")
+    end
+  end
```
kou, line 171: Are they the same concept tests?
Can we remove `test_unquoted_cr_rejected_when_included_in_row_separator`?

test/csv/parse/test_general.rb
```diff
@@ -138,28 +138,71 @@ def test_non_regex_edge_cases
...
-line,5,jkl
-    CSV
+  def test_liberal_parsing_with_unquoted_cr_and_custom_row_separator
+    data = "field1,field\rwith\rcr,field3|row2,data,here|"
+    expected = [
+      ["field1", "field\rwith\rcr", "field3"],
+      ["row2", "data", "here"]
+    ]
+    assert_equal(expected, CSV.parse(data, row_sep: "|", liberal_parsing: true))
+  end
```
kou, line 180: Can we move this to `test/csv/parse/test_liberal_parsing` and remove the `liberal_parsing_with_` part from the test name?

test/csv/parse/test_general.rb
```diff
@@ -138,28 +138,55 @@ def test_non_regex_edge_cases
...
+  def test_unquoted_cr_with_custom_row_separator
+    data = "field1,field\rwith\rcr,field3|row2,data,here|"
+    expected = [
+      ["field1", "field\rwith\rcr", "field3"],
+      ["row2", "data", "here"]
+    ]
+    assert_equal(expected, CSV.parse(data, row_sep: "|"))
+  end
+
+  def test_unquoted_cr_rejected_when_included_in_row_separator
```
kou, line 159: We can use the same naming rule for parse error case.

```suggestion
  def test_unquoted_cr_with_crlf_row_separator
```

test/csv/parse/test_general.rb
```diff
@@ -138,28 +138,55 @@ def test_non_regex_edge_cases
...
+    data = "field1,field\r2,field3\r\nrow2,data,here\r\n"
+    assert_raise(CSV::MalformedCSVError) do
```
kou, line 161: Could you also check the error message like other tests?

```suggestion
    message = "..."
    assert_raise(CSV::MalformedCSVError.new(message, 1)) do
```

lib/csv/parser.rb
```diff
@@ -675,7 +675,9 @@ def prepare_quoted
     def prepare_unquoted
       return if @quote_character.nil?
 
-      no_unquoted_values = "\r\n".encode(@encoding)
+      # Only exclude characters that are actually part of the row separator
+      # instead of hardcoding "\r\n"
```
kou, line 679: It seems that we can remove this comment.
I feel that it's useful for commit message (the PR description in this repository) because it describes why we do this change but it may not be useful for readers of new code. (Nobody will not try using `"\r\n"` here.)

PR #348, opened 2025-07-02
Increase speed of Table#to_csv

wishdev:
If we set the encoding when we call Table#to_csv we do not need to go row by row to determine the encoding to use. This allows the use of CSV#generate_lines as a faster exporter.

Add a single_encoding option to Table#to_csv to use the encoding of the first row of the table as the encoding.

### DIFF
--- a/test/csv/test_table.rb
+++ b/test/csv/test_table.rb
@@ -378,7 +378,7 @@ A,B,C
               CSV::Row.new(%w{A}, ["\x00\xac"]) ]
     table = CSV::Table.new(rows)
 
-    assert_equal('UTF-8', table.to_csv(encoding: 'UTF-8').encoding.to_s)
+    assert_equal(Encoding::UTF_8, table.to_csv(encoding: 'UTF-8').encoding)
     assert_raises(Encoding::CompatibilityError) {table.to_csv}
   end
 


kou:
Hmm. I'm not sure whether this approach is optimal.

Could you also add a benchmark to https://github.com/ruby/csv/tree/main/benchmark ? Let's discuss which approach is better with numbers.

wishdev:
Benchmark added

At 100 rows I get about a 5.8x improvement
At 1000 rows I get about a 6.7x improvement

wishdev:
So the first test example has a string that is valid as either ASCII-8BIT or UTF-8. Currently `table.to_csv` fails to process that table because it is 2 encodings that are not compatible. What does `detect_encoding_with_full_row_scan` do with that? Is it a fast fail? If 2 compatible encodings exist within the file which one wins? Currently it's the first encoding on a line that gets chosen - do we now scan each column?

I don't know enough about encodings to figure out what a default option should be. This was crafted to offer advantages without interfering for a second with existing code. I'm not comfortable for `detect_encoding_with_full_row_scan` because I'm not an expert in the edge cases.

Would there be objections to at least moving the row encoding detection code into something like `CSV::detect_encoding`? Fast path for `options[:encoding]`seems to not be an issue That combination would at least allows for something like `single_encoding` to be placed in a prepended module if desired by the end user (i.e. me) fairly straight forwardly.

wishdev:
I've reworked it to remove the single_encoding option and move the row encoding code into a private method of CSV (to allow my module prepending to work).

As I said I'm not comfortable modifying the existing path when the encoding is not provided. That can be left for others to investigate. This certainly does not to prevent any further enhancements.

test/csv/test_table.rb
```diff
@@ -373,6 +373,15 @@ def test_to_csv_limit_negative_over
     CSV
   end
 
+  def test_to_csv_encoding
+    rows  = [ CSV::Row.new(%w{A}, ["\x00\xac".force_encoding("ASCII-8BIT")]),
+              CSV::Row.new(%w{A}, ["\x00\xac"]) ]
+    table = CSV::Table.new(rows)
+
+    assert_equal('UTF-8', table.to_csv(encoding: 'UTF-8').encoding.to_s)
```
kou, line 381:
```suggestion
    assert_equal(Encoding::UTF_8, table.to_csv(encoding: 'UTF-8').encoding)
```

kou (commented):
Could you rebase on main to fix the `/opt/hostedtoolcache/Ruby/3.4.4/x64/lib/ruby/3.4.0/rubygems/dependency.rb:303:in 'Gem::Dependency#to_specs': Could not find 'csv' (= 3.3.0) - did find: [csv-3.3.2,csv-3.0.2,csv-3.0.1] (Gem::MissingSpecVersionError)` in our benchmark CI?

Can we discuss the `row_encoding` extraction as a separated issue? I'm not sure whether it's a good approach for now...

test/csv/test_table.rb
wishdev: Done!

wishdev:
Rebased to the latest main

wishdev:
I'm more than happy to discuss anything at all. However, as I've been pretty clear on, I'm not comfortable changing the default as I don't know enough about encoding to analyze that type of change.

I think what is currently here is a great starting point for any further changes along those lines in the future........

PR #349, opened 2025-07-04
ci benchmark: ensure installing old versions

kou:
Benchmark CI is failing because csv 3.3.0 isn't installed:

https://github.com/ruby/csv/actions/runs/15481241182/job/43587350229#step:5:10

```text
/opt/hostedtoolcache/Ruby/3.4.4/x64/lib/ruby/3.4.0/rubygems/dependency.rb:303:in 'Gem::Dependency#to_specs': Could not find 'csv' (= 3.3.0) - did find: [csv-3.3.2,csv-3.0.2,csv-3.0.1] (Gem::MissingSpecVersionError)
```

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -15,7 +15,7 @@ jobs:
           - ubuntu-latest
           - windows-latest
     runs-on: ${{ matrix.runs-on }}
-    timeout-minutes: 10
+    timeout-minutes: 15
     steps:
       - uses: actions/checkout@v4
       - uses: ruby/setup-ruby@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/benchmark.yml
```diff
@@ -15,7 +15,7 @@ jobs:
           - ubuntu-latest
           - windows-latest
     runs-on: ${{ matrix.runs-on }}
-    timeout-minutes: 10
+    timeout-minutes: 15
```
kou, line 18: It seems that 10min isn't enough on Windows based on the results on my fork: https://github.com/kou/csv/actions/runs/16061114221

PR #355, opened 2026-03-02
Fix frozen string literal warning

kha-wogi:
This PR fixes the frozen string literal warning

```
/.asdf/installs/ruby/3.4.7/lib/ruby/gems/3.4.0/gems/csv-3.3.0/lib/csv/writer.rb:56: warning: literal string will be frozen in the future (run with --debug-frozen-string-literal for more information)
```


kou:
Could you provide a script that reproduces this?

It seems that an application provides the frozen string.

PR #356, opened 2026-04-29
Support block in CSV::Row#to_h

tsymbalenkovlad:
It will be consistent with Ruby's method, otherwise it has to be like `row.entries.to_h { ... }`

### DIFF
--- a/lib/csv/row.rb
+++ b/lib/csv/row.rb
@@ -652,7 +652,8 @@ class CSV
     #   row.to_h # => {"Name"=>"Foo"}
     def to_h
       hash = {}
-      each do |key, value|
+      each do |key, _value|
+        value = self[key]
         key, value = yield(key, value) if block_given?
 
         hash[key] = value unless hash.key?(key)
--- a/lib/csv/row.rb
+++ b/lib/csv/row.rb
@@ -653,12 +653,9 @@ class CSV
     def to_h
       hash = {}
       each do |key, value|
-        new_key, new_value = if block_given?
-          yield(key, value)
-        else
-          [key, value]
-        end
-        hash[new_key] = new_value unless hash.key?(new_key)
+        key, value = yield(key, value) if block_given?
+
+        hash[key] = value unless hash.key?(key)
       end
       hash
     end
--- a/lib/csv/row.rb
+++ b/lib/csv/row.rb
@@ -637,6 +637,7 @@ class CSV
 
     # :call-seq:
     #   row.to_h -> hash
+    #   row.to_h {|key, value| ... } -> hash
     #
     # Returns the new \Hash formed by adding each header-value pair in +self+
     # as a key-value pair in the \Hash.
--- a/lib/csv/row.rb
+++ b/lib/csv/row.rb
@@ -650,13 +650,21 @@ class CSV
     #   table = CSV.parse(source, headers: true)
     #   row = table[0]
     #   row.to_h # => {"Name"=>"Foo"}
+    #
+    # If a block is given, will call it with (key, value) arguments and use result as a hash entry:
+    #   source = "Name,Value\nfoo,1\nbar,2\nbaz,3\n"
+    #   table = CSV.parse(source, headers: true)
+    #   row = table[0]
+    #   row.to_h { |key, value| [key, value.to_i * 2] } # => {"Name"=>"foo", "Value"=>2}
     def to_h
       hash = {}
       each do |key, _value|
+        next if hash.key?(key)
+
         value = self[key]
         key, value = yield(key, value) if block_given?
 
-        hash[key] = value unless hash.key?(key)
+        hash[key] = value
       end
       hash
     end
--- a/lib/csv/row.rb
+++ b/lib/csv/row.rb
@@ -652,8 +652,13 @@ class CSV
     #   row.to_h # => {"Name"=>"Foo"}
     def to_h
       hash = {}
-      each do |key, _value|
-        hash[key] = self[key] unless hash.key?(key)
+      each do |key, value|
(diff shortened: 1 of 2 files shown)


lib/csv/row.rb
```diff
@@ -652,8 +652,13 @@ def ==(other)
     #   row.to_h # => {"Name"=>"Foo"}
     def to_h
       hash = {}
-      each do |key, _value|
-        hash[key] = self[key] unless hash.key?(key)
+      each do |key, value|
+        new_key, new_value = if block_given?
+          yield(key, value)
+        else
+          [key, value]
```
kou: We don't want to create a needless temporary `Array`.

Could you use

```ruby
if block_given?
  each do |key, value|
    key, value = yield(key, value)
    # ...
  end
else
  each do |key, value|
    # ...
  end
end
```

or

```ruby
each do |key, value|
  key, value = yield(key, value) if block_given?
  # ...
end
```

?

test/csv/test_row.rb
```diff
@@ -339,6 +339,9 @@ def test_to_hash
       assert_predicate(string_key, :frozen?)
       assert_same(string_key, @row.headers[h])
     end
+    row2 =  CSV::Row.new(%w{A B C}, [1, 2, 3])
+    hash2 = row2.to_hash { |k, v| [k, v ** 2] }
+    assert_equal({"A" => 1, "B" => 4, "C" => 9}, hash2)
```
kou: Could you create a separated test?

lib/csv/row.rb
tsymbalenkovlad: Thanks, that's a good one

lib/csv/row.rb
```diff
@@ -652,8 +652,10 @@ def ==(other)
     #   row.to_h # => {"Name"=>"Foo"}
     def to_h
       hash = {}
-      each do |key, _value|
-        hash[key] = self[key] unless hash.key?(key)
+      each do |key, value|
+        key, value = yield(key, value) if block_given?
```
kou: Ah, sorry. We need to call `self[key]`:

```suggestion
      each do |key, _value|
        value = self[key]
        key, value = yield(key, value) if block_given?
```

test/csv/test_row.rb
```diff
@@ -341,6 +341,12 @@ def test_to_hash
     end
   end
 
+  def test_to_hash_with_block
+    row =  CSV::Row.new(%w{A B C}, [1, 2, 3])
```
kou:
```suggestion
    row = CSV::Row.new(%w{A B C}, [1, 2, 3])
```

lib/csv/row.rb
```diff
@@ -650,10 +651,21 @@ def ==(other)
...
+    # If a block is given, will call it with (key, value) arguments and use result as a hash entry:
+    #   source = "Name,Value\nfoo,1\nbar,2\nbaz,3\n"
+    #   table = CSV.parse(source, headers: true)
+    #   row = table[0]
+    #   row.to_h { |key, value| [key, value.to_i * 2] } # => {"Name"=>"foo", "Value"=>2}
     def to_h
       hash = {}
       each do |key, _value|
-        hash[key] = self[key] unless hash.key?(key)
+        next if hash.key?(key)
```
kou: `key` may be changed by `yield(key, value)`. So we can't do this if `block_given?`.

How about using separated `each`?

```ruby
if block_given?
  each do |key, _value|
    key, value = yield(key, self[key])
    hash[key] = value unless hash.key?(key)
  end
else
  each do |key, _value|
    hash[key] = self[key] unless hash.key?(key)
  end
end
```

lib/csv/row.rb
```diff
@@ -650,11 +651,31 @@ def ==(other)
...
+
+      if block_given?
+        each do |key, _value|
+          result = Array.try_convert(yield(key, self[key]))
+          raise TypeError, "wrong element type #{result.class} (expected array)" if result.nil?
+          raise ArgumentError, "wrong array length (expected 2, was #{result.size})" unless result.size == 2
+
+          key, value = result
+          key.freeze if key.is_a?(String) && !key.frozen?
+          hash[key] = value unless hash.key?(key)
```
kou: Could you avoid needless freeze check?

```suggestion
          next if hash.key?(key)

          key.freeze if key.is_a?(String) && !key.frozen?
          hash[key] = value
```

kou: @tsymbalenkovlad What do you think about this?

tsymbalenkovlad: Oh sorry I didn't see code suggestion and thought it's about removing `!key.frozen?`

test/csv/test_row.rb
```diff
@@ -343,10 +343,16 @@ def test_to_hash
...
+
     new_keys_map = {"A" => "A", "B" => "B", "C" => "B"}
-    hash = row.to_hash { |k, v| [new_keys_map[k], v**2] }
-    assert_equal({"A" => 1, "B" => 4}, hash)
-    hash.each_key do |string_key|
+    hash2 = row.to_hash { |k, v| [new_keys_map[k], v**2] }
+    assert_equal({"A" => 1, "B" => 4}, hash2)
+    hash2.each_key do |string_key|
       assert_predicate(string_key, :frozen?)
     end
```
kou: Could you split this duplicated key case to a separated test?

We can merge this after it.

lib/csv/row.rb
kou: Thanks.

kou:
Thanks.

Do you want to use this feature soon? (Should we release a new version soon?)

tsymbalenkovlad:
@kou No rush, I have a patch in project

Thanks

Issue #107, opened 2019-10-03
CSV parser fails with mixed quoting and non-standard row_sep


yourpalal:
```
opts = {row_sep: "|\n", col_sep: ","}
CSV.parse(CSV.generate(opts) { |csv|
  csv << ["yes, it's true"];
  csv << [ "CSV is broken"];
  csv << ["uhoh!"]; }, opts)
```

In the example above, the parser will raise a `CSV::MalformedCSVError (Unquoted fields do not allow new line <"\n"> in line 2.)` error.

It will succeed if any of the following are done:

1. pass `force_quotes: true` to `generate`
2. add a comma in each row
3. remove the comma from the first row
4. use `row_sep: "\n"`

Therefore, the problem seems to be that a non-standard `row_sep` + lines with a field being quoted in one row and unquoted in the next.

yourpalal:
I'm using the version included in ruby 2.6.5, btw!

yourpalal:
The stack trace indicates that it's failing in `parse_quotable_robust`. I'm thinking the problem may be that in `prepare_unquoted`, the @unquoted_value regexp does not take into account the value of row_sep. Therefore, the '|' is consumed by the scanner as part of the unquoted field value. After this, the `"|\n"` no longer exists to denote a line break, and instead the \n is interpreted as the next character of this field.

kou:
Why do you want to use `"|\n"`?
Could you show a real world use case?

yourpalal:
We're using this because we're interfacing with other systems that have very poor support for CSV, and get confused by newlines in quoted fields. That said, these files are created and then transformed by our software before being passed off to the other software (various RDBMSes) so during that first, internal phase, we could just use the default row/column separators, and avoid needing to parse the non-standard format.

If you don't want to support it, that's ok but an error should probably be raised when reading with non-standard row/col separators.

kou:
Could you show the real CSV that has problem?
The CSV does NOT use `"|\n"` as row separator, right?

yourpalal:
The CSV was using "|\n" as a row separator. 

This CSV is used for an export/import process that involves creating CSV files, transforming them, and then finally importing them into MySQL, Postgres, Oracle, etc. We use a "|\n" separator in the final CSV files to overcome very poor CSV support in some of these databases.

[...]

PR #365, opened 2026-07-28
Ignore quotes when detecting the row separator

gaoflow:
`:auto` row-separator detection scans the sample for the first CR/LF without tracking quote state, so a CR inside a quoted field is mistaken for the row separator. CSV therefore cannot parse its own `generate` output whenever a field contains `\r`:

```ruby
CSV.parse(CSV.generate {|c| c << ["a\rb", "x"]})
# => [["a\rb", "x\n"]]   # wrong: the \n was swallowed into the last field
CSV.parse(CSV.generate {|c| c << ["a\r\nb", "x"]})
# => CSV::MalformedCSVError: Unquoted fields do not allow new line
CSV.parse(CSV.generate {|c| c << ["\r"]})
# => CSV::MalformedCSVError: Any value after quoted field isn't allowed
```

All three parse correctly with an explicit `row_sep: "\n"`, which isolates the fault to `:auto` detection. `detect_row_separator` now skips CR/LF that fall inside a quoted field. An unquoted CR is still honored as a separator, so old Mac line endings keep working, and explicit `row_sep` paths are untouched.

Added round-trip and detection tests in `test/csv/parse/test_row_separator.rb`; the full suite stays green (527 tests, 4050 assertions). This is distinct from #107, which is about an explicit non-standard `row_sep`.


lib/csv/parser.rb
```diff
@@ -754,6 +754,21 @@ def detect_row_separator(sample, cr, lf)
       end
     end
 
+    # Index of the first +target+ that is outside a quoted field, so a CR or
+    # LF inside quotes is not mistaken for the row separator.
+    def unquoted_index(sample, target)
```
kou: * Could you rename `unquoted_index` to `find_unquouted_string_index` or something? I feel that no verb method name is strange for this because the receiver is `Parser` not `String`.
* Could you rename `sample` to `text` or `string` because sample is strange in this context?

gaoflow: Done — renamed to `find_unquoted_string_index` and the parameter to `text` (5951706). Row-separator tests pass.

